PR #669, opened 2014-02-26
Added styles for dropdown error state

musatov:
Styles for dropdown error state work for "selection" type of dopdown. Error class can be applied to dropdown, form field which contains dropdown or to group of fields. Look and feel of dropdown will be the same in all cases above. Also added examples into dropdown module documentation.


mjayson29:
hi, how i can change the state of my dropdown from default to error vice versa. thanks!

sstok:
@mjayson29 https://github.com/Semantic-Org/Semantic-UI#user-support

PR #1156, opened 2014-10-09
Fixed npm - replaced hidden module with _module variable

AnsonT:
The npm packages do not work because the module exports variable is hidden by the locally defined module  variable within the functions - module.exports.settings is undefined. This fix adds a _module alias. A better but more involved fix is to not use module as local variable name, but that requires changing all the js files.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -596,8 +596,16 @@ module.exports = function(grunt) {
         options: {
           patterns: [
             {
-              match: /\$.fn.\w+/g,
-              replacement: 'module.exports'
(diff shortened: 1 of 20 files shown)


hungryzi:
Why is `transition` not exported as a module? :-/

jlukic:
@hungryzi https://github.com/Semantic-Org/UI-Transition

davide:
@hungryzi pull request 2816 (linked above) will correctly export transition as a module. Not sure when the next npm release is due though.

PR #1552, opened 2015-01-02
RTL Gulp Task

MohammadYounes:


jlukic:
Added in 7b3919f789c1103d30e74250d7bde5497a8dd0bf 

I also added  an ``rtl watch`` task. 
4fa3100c5838456893e49dba2117a8ae10873147

You can run tasks with spaces in gulp like this 

```
gulp "watch rtl"
gulp "build rtl"
```

I will also add a value in ``semantic.json`` ``rtl: true`` which will run `rtl` tasks from ``gulp watch` and `gulp build`

jlukic:
You can now run rtl tasks with `gulp build` and `gulp watch` when `semantic.json` has `rtl: true` ebda601

jlukic:
I think this should be everything. Just need to confirm RTLCSS changes are correctly adjusting styles.

MohammadYounes:
RTL task is based on release files, so its required to run after normal build.  also this https://github.com/Semantic-Org/Semantic-UI/commit/4fa3100c5838456893e49dba2117a8ae10873147#commitcomment-9190308

jlukic:
Adding with `1.7.0`

PR #1553, opened 2015-01-02
Fix relative paths on windows

MohammadYounes:
Ensures consistent URLs such as  `url("../themes/default/assets/fonts/icons.eot");` instead of `url("..\themes/default/assets/fonts/icons.eot");`


jlukic:
Ah I see, these are URL paths that are put into css not filesystem paths. I get it now.

PR #1557, opened 2015-01-03
Bug fix #1554, modified .ui.segment.attached to fix content pane bleeding

### LINKED ISSUE #1554: Tabular menu: content has longer width than the menu
See: http://jsfiddle.net/6murnr7n/20/

UiCandy:
Minor fix for issue #1554, as described by @mxth, here: http://jsfiddle.net/6murnr7n/20/

Changing `.ui.segment.attached { width: calc(100% + 2px) }` to `.ui.segment.attached { width:100% }` thus ensuring there's no bleed on the content pane and it aligns itself to the nav perfectly.

Cheers,
Rishav S

### DIFF
--- a/dist/semantic.css
+++ b/dist/semantic.css
@@ -12293,10 +12293,8 @@ ol.ui.horizontal.list li:before,
   top: 0px;
   bottom: 0px;
   margin: 0em -1px;
-  width: -webkit-calc(100% +  2px );
-  width: calc(100% +  2px );
-  max-width: -webkit-calc(100% +  2px );
-  max-width: calc(100% +  2px );
+  width: 100%;  
+  max-width: 100%;
   border-radius: 0px;
   box-shadow: none;
   border: 1px solid #d4d4d5;


MohammadYounes:
You should not be changing `dist/semantic.css`, this file is the result of building semantic source files.

The source file for segment element is located [here](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/elements/segment.less#L553) and the variable for the *default* theme is located [here](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/elements/segment.variables#L77)

UiCandy:
Thanks for that Mohammad,
I'll submit an updated version tomorrow. Cheers!

MohammadYounes:
I think you should revise it again, as your update fixes the issue with the tabular menu, but it does not work well with other attached segments, see http://jsfiddle.net/j1vdakx5/

UiCandy:
Thanks for simulating that behavior, I'm going to look at it with a greater level of detail and push it.

Cheers,
Rishav S

jlukic:
The difficulty is aligning different ui elements that have borders created using `box-shadow` and `border`. The adjacent sibling selector ``.ui.tabular.menu ~ .segment`` will also match other segments later in content flow, as long as they are on the same level. So its a bit course and can also cause issues.

I'm not sure of any good way to go forward other than requiring a special class on attached tab segments.

jlukic:
Continued in #1587

Issue #1385, opened 2014-12-05
Retooling Build Tools


jlukic:
I spent some extra time in the park this morning pondering things in #1350 

I've come up with some thoughts of how I think we should move this forward to improve workflow for ``npm`` and ``bower``. 

For reference "pm folder" in this discussion is referring to the local *package managed folder*  i.e. ``node_modules/semantic-ui`` or ``bower_components/semantic-ui``.

##### Nothing that should be permanent should exist in a PM Folder

Maintainers of both package managers have made it explicit that these folders may be wiped during updates, and generally no one wants to ``cd`` into their components folder. 

This means that an install should generally not change files inside of its own directory, because it can't expect them to exist on update. If it does has to modify a file in its own directory, than it must make sure that it is trivial to regenerate this file on update.

##### LESS ``import`` must be hard-coded inside definitions

Currently LESS theming and path variables are set in a ``theme.config`` file which is expected to exist in PM folder ``src/``. This is because any ``import`` paths in a definitions must be hard-coded to something because definition files can't be modified after install or will risk losing changes (see issue above).

If LESS theming variables were added as run-time variables pulled in from ``semantic.json``, then only our build tools would work with our `less` files. Which is a functional solution, but would not allow the ``less`` source to be compiled without our build tools. This might alienate developers.

### Proposal Summary

* Gulp install will check if installed with ``npm`` or ``bower``. 
* If using PM it will suggest user paths in project root. 
* Install will also create a gulpfile at project root or add tasks to an existing one. 
* Install will create paths if not currently available
* Install will modify PM folder file ``src/theme.config`` to a single line file that imports a user's real ``theme.config`` located outside of pm folder.
* This file will be considered a temporary file (that could be lost on update/upgrade) and should be automatically regenerated if not found by a backup value stored in ``semantic.json``

##### 1) Install tools should recursively search for a ``semantic.json`` across parent directories. 

This would mean having to modify the project's gulpfile to walk recursively through each parent directory until it hit a ``semantic.json``.

Having a ``semantic.json`` at project root would be similar to having a ``gulpfile`` ``gruntfile`` or ``package.json`` or ``karma.json`` which is a familiar idea for developers, and reduce the barrier for entry for new developers.

##### 2) Theme folder should include two paths. One for default theme inside PM folder, and another for user themes in project root.

This is to deal with package managed folders being wiped on update or machine change. We will need to modify ``theme.less`` to use separate paths for default theme and user themes.

[...]

PR #1602, opened 2015-01-11
Fix i.icon selector with attribute

lolychee:


victorpolko:
This commit is very important. Rails app's assets fail to compile with error:
```bash
Sass::SyntaxError: Invalid CSS after "...="top aligned"]": expected "{", was "i.icon,.ui.list..."

"i.icon" may only be used at the beginning of a compound selector.
```
We use bower so it's uneasy to rewrite this file manually.

Thank you, @lolychee

jlukic:
I'll patch and release this today.

`1.7.0` will come tomorrow with new features.

jlukic:
Please be careful with the current state of SUI in bower #1385

jlukic:
Released.

victorpolko:
@jlukic, thank you very much!
Unfortunately, I've ran into Asset Pipeline troubles with CSS (direct links to fonts in semantic.css, while AP uses fingerprinting) so that I fell back from Bower to Semantic-UI gem for now. Of course, I don't like it, but I need a fast working result now, so I decided to do so. I will watch your repo and will go back to Bower as soon as it will be working well.
Thanks again!

Issue #1601, opened 2015-01-11
[Dropdown] Search Selection Produces Error with Numerical Values


zelenin:
```
<select>
<option value="1">Name A</option>
<option value="2">Name B</option>
</select>
```
if value is numerical (common case) search not works with error:

```
TypeError: value.match is not a function	

if( text.match(exactRegExp) || value.match(exactRegExp) ) {
	
semantic.js (line 3395, col 45)
```
I think it should cast value to string in search.
version 1.6.2

jlukic:
I'm unable to reproduce any issues with the regexp with your code example. 
http://jsfiddle.net/sg482x91/

zelenin:
@jlukic http://jsfiddle.net/sg482x91/1/ enter A to search
http://i.imgur.com/fVERfaM.png 
Firefox 34

jlukic:
Ah, with search selection. Thanks

jlukic:
While I was in there, I noticed regExp were also not escaping special characters. Fixed both bugs in `next`

PR #1620, opened 2015-01-13
Fixed issue #1601

alufers:
Search Selection Error with Numerical Values


jlukic:
Sorry about that was working on the same thing at same time. Details in #1601

Issue #1621, opened 2015-01-13
Extending transitions


ph7vc:
I like having transitions shipped with the framework, however it feels somewhat limited.

Is there a reason that there isn't support for something like "slide in left"? There only appears to be up and down at the moment.

I was trying to look at how I could extend it; and it looks like the CSS for the transitions are in the `src/themes/default/modules/transition.overrides` file.

I guess my question is; how does `transition.js` know what transitions are valid? Is it as simple as adding the appropriate CSS classes?
 
Thanks!

jlukic:
Its really simple to extend. 

Transition creates a fake element tests if an animation exists (after adding the appropriate class names), then caches the results for future animations. It also uses this fake element to determine whether the animation is static, or has both an `in` and `out` animation. 

You can extend transition relatively easily. I would be willing to include pull requests that add new animations.  Any animation that transitions in and out just needs one rule for each. 

So `slide left` would need `slide left in` and `slide left out`.

For static animations like bounce etc. You can just have ``bounce`` as class name.

Since the definitions are arbitrary they all get put into the default themes override file.

PR #1623, opened 2015-01-14
Transitions extension

ph7vc:
Added / modified default transitions. 
Mentioned in #1621 

## Appearance
* up down left right with entrance and exit variations for all

**Fade:** 
* Has no direction by default, simply fades opacity

**Bounce:**
* Defaults to bouncing along Z-axis
* Modifed version from: https://github.com/daneden/animate.css/tree/master/source/bouncing_entrances

**Slide:**
* Defaults to sliding from the top

## Emphasis
**Rubberband:**
* Copied directly from: https://github.com/daneden/animate.css/blob/master/source/attention_seekers/rubberBand.css

Demo of additions: http://patcave.info/semanticui-transitions/

### DIFF
--- a/src/themes/default/modules/transition.overrides
+++ b/src/themes/default/modules/transition.overrides
@@ -196,96 +196,6 @@
 }
 
 /*--------------
-      Fades
----------------*/
-
-.fade.transition.in {
-  animation-name: fadeIn;
-}
-.fade.transition.out {
-  animation-name: fadeOut;
-}
-
-.fade.up.transition.in {
-  animation-name: fadeUpIn;
-}
-.fade.up.transition.out {
-  animation-name: fadeUpOut;
-}
-
-.fade.down.transition.in {
-  animation-name: fadeDownIn;
-}
-.fade.down.transition.out {
-  animation-name: fadeDownOut;
-}
-
-
-/* Fade */
-@keyframes fadeIn {
-  0% {
-    opacity: 0;
-  }
-  100% {
-    opacity: 1;
-  }
-}
-
-@keyframes fadeOut {
-  0% {
-    opacity: 1;
-  }
-  100% {
-    opacity: 0;
-  }
-}
-
-/* Fade Up */
-@keyframes fadeUpIn {
-  0% {
-    opacity: 0;
-    transform: translateY(10%);
-  }
-  100% {
-    opacity: 1;
-    transform: translateY(0%);
-  }
-}
-
-@keyframes fadeUpOut {
-  0% {
-    opacity: 1;
-    transform: translateY(0%);
-  }
-  100% {
-    opacity: 0;
-    transform: translateY(10%);
-  }
-}
-
-/* Fade Down */
-@keyframes fadeDownIn {
-  0% {
-    opacity: 0;
-    transform: translateY(-10%);
-  }
-  100% {
-    opacity: 1;
-    transform: translateY(0%);
-  }
-}
-@keyframes fadeDownOut {
-  0% {
-    opacity: 1;
-    transform: translateY(0%);
-  }
-  100% {
-    opacity: 0;
-    transform: translateY(-10%);
-  }
-}
-
-/*--------------
       Scale
 ---------------*/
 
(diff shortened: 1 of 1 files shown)


Lisa-Williams:
You better check the console log after clicking some "In" and "Out" buttons in your demo.

jlukic:
@Lisa-Williams It just looks like this happens when no animation is selected, doesn't appear to be an issue with contributions just sanity-checking user input.

jlukic:
These are great. Thanks for the hard work and terrific demo.

**Bounce**

We have some naming issues since bounce is already a static animation in existing versions
http://semantic-ui.com/modules/transition.html#bounce

Since direction is determined by current visibility state, there cannot exist both a static animation and transition with the same name. Renaming `bounce` would create *breaking changes* which should, with SemVer, be only introduced in major releases, i.e. `2.x`

Bounce is probably the most appropriate name, but we will have to find another. Perhaps "swing" "oscillate", "rebound", "jump", or do you have any suggestions?

**Rubberband** 

All other transitions use verbs. Are you fine with changing this to something like "stretch", or "bend"?

jlukic:
I'll have to fit this into `1.8.0`

ph7vc:
For **Rubberband**, we could just call it **Jiggle** perhaps?

As far as **Bounce** is concerned, I really have no idea what would be appropriate since this animation seems to combine a few actions. 

We should also be careful since like you said, renaming animations will be very difficult. If in the future someone introduces an animation that is much better suited for the name we have for another, what then? There's seems to be a limited number of words to describe animations.

ph7vc:
Also, the in/out direction is something that tripped me up. If we have an animation like *"fade right"*, does it fade in from the right, then fade out to the right? Or should it return to the direction it came from?

Lisa-Williams:
@jlukic You are right. I just checked again. The dimmer did it's job almost too good.
These are indeed a great addition. Thanks @ph7vc .

Lisa-Williams:
Forgot.
"Bounce Up" then "In" flashes a vertical scrollbar.
So does "Bounce Down" then "Out".
Will this be just in the demo?

ph7vc:
@Lisa-Williams I didn't even notice that. It looks like it happens because of where the animation starts. It does it on http://daneden.github.io/animate.css/ as well, which is where I got the animation from. I'll see if I cant fix it

Lisa-Williams:
Well if you could. Because I would love to use these.

Lisa-Williams:
@ph7vc I found a workaround.
Vertical scrollbar means: page gets taller.
Specify ```overflow-y: hidden``` on the ```html``` element and it hides the vertical scrollbar.
But this should only happen for the duration of the animation. 0% and 100% in the keyframe animation.
Possibly not the best resolution though.

ph7vc:
@Lisa-Williams the problem is that this also seems to happen vertically as well. Hiding overflow can cause problems for people who have websites that scroll horizontally or vertically. I don't think that would be a viable solution. The only think I can think of is to reduce the distance it comes in from, 3000px to something smaller

jlukic:
This is one of the limitations with global animations, `scale in` suffers the same fate with modal, `10%` is pretty subtle on small elements, but on massive modals it can be a huge jump.

Lisa-Williams:
@ph7vc If the overflow-y: hidden is set at 0% and removed (overflow-y: visible or initial) at 100% it is almost unnoticeable. Unless the animation would take multiple seconds. But yes, it is far from ideal.

@jlukic What about ```Boink```, ```Boing```, ```Bam Bam``` to replace ```Bounce```?

ph7vc:
I added another transition called "Swing".

For some reason, on the demo it doesn't seem to pick up the in and out when the select box is changed.

Demo: http://patcave.info/semanticui-transitions/

jlukic:
Thanks again. Will make all this into next week's release. `1.8.0`

jlukic:
With regards to transitioning out, they should return to the same point which the animation began. This seems to be what you have already.

jlukic:
This is all merged in. 

I've changed transition module to not set `animation-duration` by default. You can still set transition duration in javascript, but it will obey css defaults if no duration is specified. This should make it easier for people who want to just add `animation-duration` in their custom css. 

In addition I've modified `transition.overrides` to give some animations additional time. For example, `fly`, `drop`, and `swing`, all need extra animation duration to account for their telegraphed easing.

PR #1627, opened 2015-01-14
Dropdown module (restore)

mktm:
This fix restores the input class "default" as well as removing the
selected item from the list in case the default value is empty (used as
placeholder).

Issue: http://jsfiddle.net/mktm/tmadsvb6/


jlukic:
Investigating..

jlukic:
Checks out, added an example to docs using your fiddle

PR #1629, opened 2015-01-14
Feature: Form reset

mktm:
Implementing a reset functionality to form module

I implemented this instead of using form native due to Semantic modules, such as checkbox and dropdown (select), plus to avoid having to do "return false" on submitting the form via Ajax.

As instance of module occurs, it will set a data value (defaultValue) to all field identifiers passed through.
As reset is called, it will look these fields and set their value to their respective defaults.
May need a bit more polishing, but works as expected.

Manual call: $('.ui.form').form('reset');

Working example: http://jsfiddle.net/mktm/y5zopezh/


mktm:
Sorry I requested the pull into master.

PR #1630, opened 2015-01-14
Form Reset + Serialize

mktm:
Implementing a reset functionality to form module

I implemented this instead of using form native due to Semantic modules, such as checkbox and dropdown (select), plus to avoid having to do "return false" on submitting the form via Ajax.

As instance of module occurs, it will set a data value (defaultValue) to all field identifiers passed through.
As reset is called, it will look these fields and set their value to their respective defaults.
May need a bit more polishing, but works as expected.

Manual call: $('.ui.form').form('reset');

Working example: http://jsfiddle.net/mktm/y5zopezh/

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -163,6 +163,8 @@ $.fn.form = function(fields, parameters) {
 
                switch (type) {
                  case 'hidden':
+                 case 'select-one':
+                 case 'select-multiple':
                    $parent.is('.ui.dropdown') && $parent.dropdown('restore defaults');
                    break;
                  case 'checkbox':
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -60,7 +60,7 @@ $.fn.form = function(fields, parameters) {
         initialize: function() {
           module.verbose('Initializing form validation', $module, validation, settings);
           module.bindEvents();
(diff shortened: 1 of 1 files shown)


mktm:
I have added a serialize function for the module.

Example: http://jsfiddle.net/y5zopezh/1/

jlukic:
Nice! I'll take a look at this for `1.8.0`

mktm:
@jlukic I am just implementing another function and changing the location of some of the old ones:
 setDefaults to set.field.defaults

As well as a set.field.values function to set input/dropdown values by passing through an object with key:value
I had used this to do a CRUD via Ajax, and not having to go through each field on the form and set them one by one. Hopefully makes use for others.

I believe tonight I will be pushing this.

jlukic:
Thanks for the update. I'll review tomorrow.

jlukic:
Err today.. delayed

jlukic:
All merged in and ready to go. Will add a version of your example to docs and give you credit in release notes.

jlukic:
Rewriting `serialize` as `get value` and `get values`. More natural with write API, and less jargon.

jlukic:
Modified to `get/read value(s)` so as not to disrupt `get.field` which is already using the name to grab field elements.

jlukic:
Added ``clear` and dropdown `clear`

PR #1632, opened 2015-01-15
Show filtered menu if search input placed in menu.

PSyton:


jlukic:
Can you explain in more detail what this change does?

jlukic:
Thanks I understand now

jlukic:
This makes more sense than what was previously in there. Good work

Issue #1633, opened 2015-01-15
Semantic build tools creates rtl css even if RTL was set to no


DHNCarlos:
When building the dist for the new 1.7.0 Semantic,during the setup it asks if you use an RTL language. No matter if I choose No or Yes, it only creates the rtl.css files.

DHNCarlos:
I also noticed that when doing an Express install and disabling some components, they are still in the compiled css files

e.g. I didn't want the Video component but after compiling, the Video component wa still in the JS file and also the styling that goes along with it in the CSS file.

jlukic:
Will fix in patch, debug code left in. 41b0c8b093b278f4d4edb9fc4725f378b16f3e8a

jlukic:
Released in `1.7.1` apologies about that. Pretty frustratingly stupid of a mistake.

DHNCarlos:
@jlukic No worries mate. Happens to the best of us :)

DHNCarlos:
@jlukic Sorry for yet another post but I think you didn't completely fix it. When going through your `tasks/questions.js` file and the section where you have the RTL language question, the values are `yes` and `no`. Now these values get defined to the `semantic.json` file as a string `"no"`. Now the issue is that you are checking `if(config.rtl)` and since it is a non-empty string, it will evaluate to `true`, no matter which value.

My suggestion would be that line 867 in `gulpfile.js` should be replaced with this:

```javascript
json.rtl = answers.rtl == "yes" ? true : false;
```

After changing to that, my gulp parses the RTL perfectly.

jlukic:
`1.9.0` is now published. I encourage you to check it out @DHNCarlos

PR #1640, opened 2015-01-15
Properly parse RTL question during gulp install

DHNCarlos:
This is a fix for the issue #1633 where the Semantic UI build tools parse the RTL question anwser value wrong.


jlukic:
Thanks

PR #1645, opened 2015-01-16
Prevent default components being merged with user-defined components

DHNCarlos:
Hello,

Currently when you have only a few components defined in your `semantic.json` file, it also merges the other components on top of the selected components from defaults.

The issue with it is, if you want to compile a custom semantic build without all of the components, it's not possible. This is due to the deep copy been set to `true` for `extend`.

Setting it to `false` allows you to compile a semantic build with only the components you want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-dist/
 node_modules/
 test/coverage/
 coverage/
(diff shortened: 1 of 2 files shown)


jlukic:
I just need to verify this doesn't effect any other nested objects that need extend.

jlukic:
The PR has some extra commits, manually merging in fix. `1.7.3`

PR #1689, opened 2015-01-22
Put popups over modals so modals can have popups.

mcantor:
Without this, modals cannot have popups because they will be obfuscated due to popups currently having a `z-index` of 900, beneath the modal's z-index of `1000`.

### DIFF
--- a/src/themes/default/modules/popup.variables
+++ b/src/themes/default/modules/popup.variables
@@ -6,7 +6,7 @@
        Element
 --------------------*/
 
-@zIndex: 900;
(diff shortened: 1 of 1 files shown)


jlukic:
Already in `next` https://github.com/Semantic-Org/Semantic-UI/commit/671948c25287ea3e0d135ee4a8eccd38193e81ee

mcantor:
Oops, I didn't know to check `next`. I didn't see it in `master` so I figured it was still open! Sorry about that.

PR #1698, opened 2015-01-23
Dropdown fix + .basic.modal 

mktm:
Remove border radius from .basic.modal
Dropdown function set.selected set to also call set.value to define the value of the hidden input to the respective active item


jlukic:
`module.set.value(value)` will not work in that context, because value is not necessarily `$selectedText` value.

Will have to get value from selected value before passing to set value. Commit to follow

jlukic:
Merged in for `1.8.0` with changes noted in commits

Issue #1744, opened 2015-02-02
Loading button should be disabled


mderazon:
I think that a `loading` button should also be disabled (without additional styling). I find it a much more common use case than a loading button that can be interacted with.
Moreover, achieving both is not possible by simply setting `disabled loading` since `disabled` changes the color of the button. So if I have a colored loading button, setting `disabled` makes it grey.

If making this breaking change is impossible, then perhaps introduce a new attribute `muted` (`disabled` and `active` were taken :) ) that when added will achieve the same...

What do you think ?

ph7vc:
+1

jlukic:
When you say disabled do you mean specifically [`point-events: none;`](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/elements/button.less#L211)?

mderazon:
Yes, precisely.

ph7vc:
@jlukic: I think he means when it has class `loading`, the button should not have hover or active states as to not give the user the impression that they can still interact with the button.

mderazon:
Yes, like @ph7vc said but perhapes also not fire any mouse / click events

jlukic:
I understand. I'll take a look at loading buttons again. 

Perhaps there's some middle ground for loading thats an opacity a notch above disabled. At the very least it should be a variable.

jlukic:
Loading will no longer receive events in default theme. 

Added variables for loading opacity and pointer-events.

Available in `next`

mderazon:
Awesome, thanks :+1:

PR #1755, opened 2015-02-04
Disabling loading buttons with LESS variable

ph7vc:
References #1744

LESS has the ability to do pseudo if statements. Given this, we define a boolean for disabling the loading button state

I covered a few cases to determine how it compiles. It will tack on a selector to the normal states when this variable is set to true. It does not create any extra bloat when compiled into the `semantic.css` file

I can proceed to add all other variants (basic inverted colored, etc) if this works for you @jlukic 

Demo: http://patcave.info/semantic-fork/buttons.html

Edit: `pointer-events:none` is not supported for < IE 11, but would work for this as well. See [this](http://caniuse.com/#feat=pointer-events)

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -139,6 +139,21 @@
   color: transparent !important;
   transition: all 0s linear;
 }
+
+.ui.loading.button when (@loadingDisabled = true) {
+    /* Normal */
+    &:hover, &:active{ &:extend(.ui.button); }
+    
+    /* Colors */
+    &.blue:hover, &.blue:active { &:extend(.ui.blue.button); }
+    
+    /* Inverted */
+    &.inverted:hover, &.inverted:active { &:extend(.ui.inverted.button); }
+
+    /* Basic */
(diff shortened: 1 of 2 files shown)


jlukic:
Woah nelly no mix-ins in our less.

ph7vc:
Ah. I thought remembered you mentioning that somewhere before but I wasn't sure. Is there a particular reason for that?

It's possible to write this all out by hand, I was just looking for a simple way to get it done and be able to change it quickly.

jlukic:
The logic in css is designed to be flat to remove complexity in editing, and to allow for simpler ports from LESS to future CSS precompilers. 

This means its slightly more verbose, but It also means all styles are WYSIWYG, there's no differences between pre-compiled/compiled source besides variable substitutions.  Sometimes variables use language features to help calculate values (i.e. color functions or unit changing functions) but other than that there's very few LESS specific features.

jlukic:
Added in f1b9da3 slightly different approach

Issue #1530, opened 2014-12-29
[UI] Codeblock


p-hennessy:
In similar fashion to Github, it would be nice if there was a code embed element as part of Semantic.

I imagine it having the line number on the left hand side, as well as a title bar and probably most importantly, language specific syntax highlighting. It seems that most of the building blocks for this exist already in Semantic, except for the line number and syntax highlighting.

This is an example of what I'm talking about: 
![github embed code](https://coderwall-assets-0.s3.amazonaws.com/uploads/picture/file/1162/Screen_Shot_2013-01-20_at_17.36.23.png)

sujaysreedhar:
+1

martindale:
+1

sasha240100:
+1

Elements-:
+1

hardvain:
Any updates on this?

philrykoff:
If you would go for ice9, you would also get such features as code folding and completion. I believe, it's used by github, too.

https://ace.c9.io/#nav=production

But it's probably a 'little' heavier than hightlight.js.

MurzNN:
It is needed feature!

martindale:
Keep this issue alive!

bolorundurovj:
Please how do i highlight code with semantic ui in angular

PR #1756, opened 2015-02-05
Codeblock

ns-phennessy:
Implements #1530 

This is a code block similar to what you might see on Github or Google Code.

I tried to stay consistent to the Semantic-UI standards. I wasn't sure what category this fit under, so I made it a `view`.
I pulled styles from other places in the framework and tried to use as many global variables as possible.

It is modular meaning that parts can be taken in and out as desired.

**Features**:
- Header container
  - For things like a file name, or actions that can be done on that code
  - Is attached to the top of the codeblock
- Code container
  - Where the actual code is hosted
  - It stands separate to the gutter, allowing for clean copying
  - Setup to hide scroll on overflow (JS could limit lines shown perhaps?)
- Code gutter
  - Shows line numbers or indicative icons (such as Github's codefolding)
- Footer container
  - Intended for meta information like file size or language
  - Has a divider element

**Issues / Considerations**:
- For the moment, I have not written any Javascript to do syntax highlighting or generate the line numbers for the gutter.
- The gutter and code block MUST have the same `monospace` font, otherwise the alignment of the two will not be correct

### DIFF
--- a/src/definitions/views/codeblock.less
+++ b/src/definitions/views/codeblock.less
@@ -35,7 +35,7 @@
 /*--------------
      Header
 ---------------*/
-.ui.codeblock .header{
+.ui.codeblock .header:not(.ui){
     margin:0;
     background:@headerBackgroundColor;
     color:@headerColor;
@@ -45,7 +45,7 @@
     font-weight:bold;
 }
 
-.ui.codeblock .header:after{ 
+.ui.codeblock .header:not(.ui):after{ 
     /* Clearfix */
     display:block;
     content:' ';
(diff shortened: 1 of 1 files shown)


mishak87:
Have you considered placement of comments between lines? Like is possible to do no github for example.

ph7vc:
@mishak87: While it's certainly possible, that would be an enhancement for further down the road.

As of right now, this element lacks basic functionality like line number generation, syntax highlighting, etc.

kvetoslavnovak:
ph7vc, this is really great, thank you. Nice to see other developer helping jlukic with Semantic UI. 
Looking forward your next contributions (my tip would be date picker/calendar module :-)

ph7vc:
@kvetoslavnovak, feel free to contribute as well :smiley:

PR #1764, opened 2015-02-08
Cards Improovements: images inside .content and colors for cards

romuloctba:
Today, if you place the post content inside the `.card .content` you will have bad looking `.image`s.

This PR tries to fix this, please look, example from the docs, we just take the .image and put it inside .content:
```html
<div class="ui card">
      <div class="content">
        <a class="header" href="#">Steve Jobes</a>
        <div class="meta">
          <a class="time">Last Seen 2 days ago</a>
        </div>
        <a class="image" href="#">
        <img src="/images/avatar/large/steve.jpg">
      </a>
      </div>
    </div>
```

<caption>Before the PR:</caption>

<caption>After the PR:</caption>

### DIFF
--- a/src/definitions/views/card.less
+++ b/src/definitions/views/card.less
@@ -751,11 +751,11 @@ a.ui.card:hover,
 
 @coloredBorderSize: 2px;
 .ui.black.card:not(.inverted) {
-    box-shadow: 0px @shadowDistance 0px 0px @black;
+  box-shadow: 0px @shadowDistance 0px 0px @black;
   border: 1px solid @solidBorderColor;
 }
 .ui.blue.card:not(.inverted) {
-    box-shadow: 0px @shadowDistance 0px 0px @blue;
+   box-shadow: 0px @shadowDistance 0px 0px @blue;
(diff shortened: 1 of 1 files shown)


romuloctba:
@Anachron  thank you, corrected

romuloctba:
<3 @jlukic thank you, if it need any changes, let me know

jlukic:
Merged into `next` with some adjustments 

I've omitted inverted card variations. I think forcing all content to white with `!important` is just too restrictive, and doesnt account for links etc. 

I think although `inverted segment` exist, I would probably not advocate using colored backgrounds for cards for design reasons. I can see their use on kanban boards, but even there I think they limit legibility. 

Either way changes are in `next`, and thank you very much for contributing @romuloctba

romuloctba:
Thank YOU Jack. Great fan here. For the record, i just copied from segment
most of things, cause i didn'tg know what level of abstration could we use.
I hope I can contribute much more

2015-02-16 20:46 GMT-02:00 Jack Lukic <notifications@github.com>:

PR #1765, opened 2015-02-08
[FIX] Hide modal on click selector.close

lolychee:


jlukic:
Can you clarify what case this fixes?

lolychee:
Sorry, this PR is not really fix it.

http://jsbin.com/huvibaduwa/7/edit

The first time, it's no problem.

But, the second time,  the modal is hide when I click the label tag.

And now, I can't resolve this problem.

Do you have any idea?

jlukic:
This is very interesting, I haven't seen this bug before. I'll take a look.

jlukic:
I've figured out what your issue is. 

The `<label>` field in the first modal, is activating the `<textarea>` in the first modal causing the second modal to be closed, because you are using the same `id` tag in *both modals*. 

You can resolve this by only using one modal, or changing the `for` and `id` attribute for each modal. In HTML, id tags should be unique.

lolychee:
Thanks! I forget it's same `id`.

my solution is remove the modal on hidden.
```js
$('<div class="ui modal" />').modal({
  onHidden: function() {
    $(this).remove()
  }
})
```

Thank you very much for your help!

PR #1781, opened 2015-02-11
Correct IE9 support status

ericz:
http://cl.ly/image/0Z461V1s371S
http://cl.ly/image/2b0m2T1F3y1X
etc...

I think it is misleading to list IE 9 under browser support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,10 +92,11 @@ For more detail into how work with Semantic when building a site please [read ou
 
 * Last 2 Versions FF, Chrome, IE (aka 10+)
 * Safari 6
-* IE 9+ (Browser prefix only)
+* IE 10+
 * Android 4
 * Blackberry 10
 
+Browser prefixes are present for Internet Explorer 9, but the browser is not officially supported.
 
 ## Pull Requests
 


jlukic:
Agreed. I'll remove from prefix list as well for autoprefixer.

ericz:
@jlukic Thanks for merging.

Removing the IE9 prefixes might break for some users that are already using SUI in IE9 in some way so I would vote to save removing them for the next major version. (Though they could just build themselves with a change in the settings file)

For reference of anyone else who tries to get IE9 working here are the issues I encountered getting Semantic working in IE9:

- IE9's 4096 rules / file size limit: Use [BlessCSS](http://blesscss.com/) to work around
- Default 'easeOutExpo' transition default in some modules not available, set to 'linear' instead. I didn't spend too much time on this one -- not sure why it wasn't available.
- DOM MutationObserver is used in several modules. I used the [polyfill from Polymer](https://github.com/ericz/polymer-mutationobserver)
- Dropdown (and perhaps other modules) break when using `transition: 'none'` or falling back to jQuery animations. This is a bug not specific to IE as of 1.8.3. IE9 has not CSS3 transition support, so Semantic will correctly use the fallback however the dropdown will not show because of the bug. I worked around with this CSS:
```
.ui.dropdown.active .menu:not(.transition)
  display: block !important
  visibility: visible !important
``` 

I haven't used some of the other popular modules such as popup, sidebar, etc. but dropdown, modal, search, and accordion work fine with the above steps.

PR #1785, opened 2015-02-11
Ability to always use targets offset() for popup.

PSyton:
When we set popup option to some element and target to another element sometimes we need get targets offset instead position.

For example, I use fullCalnedar event container as a target and some preloaded div as a popup content.

### DIFF
--- a/src/definitions/modules/popup.js
+++ b/src/definitions/modules/popup.js
@@ -602,14 +602,19 @@ $.fn.popup = function(parameters) {
                 ? parseInt( window.getComputedStyle(targetElement).getPropertyValue(module.is.rtl() ? 'margin-right' : 'margin-left'), 10)
                 : 0,
 
-              target        = (settings.inline || settings.popup)
+              target        = ((settings.inline || settings.popup) && !settings.useTargetOffset)
                 ? $target.position()
-                : $target.offset(),
+                : $.isPlainObject(settings.useTargetOffset) ? settings.useTargetOffset 
+                                                            : $target.offset(),
 
               computedPosition,
               positioning,
               offstagePosition
             ;
+            if ($.isPlainObject(settings.useTargetOffset)) {
+              targetWidth = 0;
+              targetHeight = 0;
+            }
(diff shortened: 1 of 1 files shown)


schmiddd:
same for me. Although I would name your parameter rather "useTargetOffset" otherwise one might actually pass the offset as a two dimensional vector.

jlukic:
Can you provide a test case to show what this enhancement changes?

PSyton:
Without this we can't see any popups https://jsfiddle.net/ykhxtp90/1/

PSyton:
https://jsfiddle.net/ykhxtp90/2/ for some events popup position is wrong.

PSyton:
https://jsfiddle.net/ja7wfvkr/2/ another example

jlukic:
I'm really sorry about the delay. I'm looking through these now

jlukic:
I'm a bit confused, the parameter name `useTargetOffset` makes it seem like it should be a `boolean`, but the code is looking for a 2 dimensional array that will replace the popups `.offset` calculations.

Is there a simple test case with just one popup that shows this issue? I'd like to see if i can fix it by adjusting the calculations

PSyton:
This is optional. useTargetOffset as object provide setting offset manually. When it boolean - used offset from target. I don't use it but added after @schmiddd comment.
May be better to add separate parameter for override target offset.

schmiddd:
Well yes, seems there are more changes by @PSyton. Originally the flag "useTargetOffset" should only override the behavior to keep the offset. See also my fork here https://github.com/schmiddd/Semantic-UI/commit/796ca2ff35f77f2a1e92abf7a63f24480d180f6c
If one would really override the position (with a two dimensional array) then a separate paramter is needed.

jlukic:
I think the thing thats blocking me from moving forward is a standalone test-case showing a popup that is pulling the wrong target offset. I'm hoping its just a miscalculation that I can adjust in the codebase caused by unusual layout conditions. 

If this is not the case, then I will concede the issue and add in a dedicated setting for targetOffset.

PR #1792, opened 2015-02-11
Make select.ui.dropdown visible if js is not available

davialexandre:
I just noticed that all my selects with ui.dropdown were invisible when js is not available


jlukic:
That's intended, this prevents the flash of unstyled content before dropdowns are initialized.

davialexandre:
That makes sense. I am closing the PR

PR #1806, opened 2015-02-15
Treatment for name[] identifier

mktm:
form.get.values: Treatment for name[] identifiers while getting the
values

With this you can use form('get values') and inputs with name attribute that is an array return the array of values set.

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -361,22 +361,42 @@ $.fn.form = function(fields, parameters) {
                   : false
               ;
               if(name) {
-                if(isRadio) {
-                  if(isChecked) {
-                    values[name] = value;
+              	if(name.indexOf('[]') !== -1) {
+              		name = name.replace('[]', '');
+              		if(!values[name]) {
+              			values[name] = [];
+              		}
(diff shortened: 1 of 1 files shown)


mktm:
Sorry I don't know why the tabulation is always broken...
I have set on my project to 2 spaces no tabs, but still...

jlukic:
I had to rework this a bit. This would work only when no [fields] array was passed into `get fields`

I also added similar lookup for `set fields`

mktm:
Great! Although, I believe this still needs a bit of polishing.

For example, we can have fields with name like identifier[array][], identifier[array1][array2][]

Which I completely forgot about when I wrote this...

jlukic:
This was more to solve the simplest case
http://semantic-ui.com/behaviors/form.html#writing-values

mktm:
Well, it will return objects with "names[like][this]": []

With PHP at least it gets interpreted since it is sent as string to the server and it builds the correct array/object structure;

Easiest way to do this for a correct output in json format would be using eval in my pov.

I am not really in pursue of this perfect, this does what I needed server-side as I get the array as it should be.
Only client-side it is not proper format.

Issue #235, opened 2013-10-11
Semantic-UI for webdesigner


Tvmak:
Hi, thank's for SEMANTIC UI. I really love this project and I want to participate ... would you be interested in a .PSD version for UX designer ?
(sorry for my poor english)

macagoraga:
any news about psd graphic files of component?

PR #1825, opened 2015-02-19
let beforeSend stop the toggling

michele-comitini:
Don't change the state when the api is defined and the request is undefined
                this allows beforeSend to stop the toggling.

### DIFF
--- a/src/definitions/behaviors/state.js
+++ b/src/definitions/behaviors/state.js
@@ -200,8 +200,10 @@ $.fn.state = function(parameters) {
                 apiRequest = $module.api('get request');
                 if(apiRequest) {
                   module.listenTo(apiRequest);
-                  return;
                 }
+                // Don't change the state when the api is defined and the request is undefined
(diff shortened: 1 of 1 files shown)


jlukic:
This should be caught by [#235](https://github.com/michele-comitini/Semantic-UI/blob/6ec8e69339abd5c90438337914c649241f5d71e0/src/definitions/behaviors/state.js#L235)

michele-comitini:
Or take the if away:

if(apiRequest) {  module.listenTo(apiRequest);  }should be simply:
module.listenTo(apiRequest)

2015-02-19 22:04 GMT+01:00 Michele Comitini <michele.comitini@gmail.com>:

jlukic:
I've added `.api('was cancelled');` to make it easier to detect aborted requests. State module now relies on this condition.

jlukic:
Releasing with `next`

michele-comitini:
Nice! thanks

PR #1845, opened 2015-02-21
Correct gulp install script's wrench config for `src/site` merges

derekslife:
Even though the gulp install script outputs:
```Site folder exists, merging files (no overwrite) project-dir/semantic/src/site/```

The `src/site` directory is blown away (not preserving existing files) due to the `forceDelete:true` setting.

### DIFF
--- a/tasks/config/project/install.js
+++ b/tasks/config/project/install.js
@@ -752,7 +752,7 @@ module.exports = {
 
       // copy for site theme
       site: {
-        forceDelete       : true,
+        forceDelete       : false,
         excludeHiddenUnix : true,
         preserveFiles     : true
       }

### CODE REVIEW

**Comment by jlukic** [comment]
My understanding is that `forceDelete` is required to add new files to an existing directory, but `preserveFiles: true` makes sure that individual files are not overwritten. 

I'm afraid if we change `forceDelete: false` then we will not be able to add new element definitions into `site/` in existing folders.

I will need to retest this.

**Comment by derekslife** [comment]
You're correct. This PR doesn't correct the problem completely. This fix preserves the files in the `site/` dir but doesn't copy any new files into `site/` during the install script's task.

There seems to be a wrench-js issue copying into an existing directory while maintaining what is already present. The `preserveFiles: true` isn't ready before it deletes the existing directory.

In wrench v1.5.0 there was a breaking change introduced which seems to be impacting the install script. If `forceDelete: true`, it will delete the existing directory, otherwise it will `return new Error(...)`. Because wrench has `return new Error()` instead of a `throw new Error()` and the install gulp script doesn't check the return for `wrench.copyDirSyncRecursive()`, this fails silently. 

From: https://github.com/ryanmcgrath/wrench-js/blob/master/lib/wrench.js
```
exports.copyDirSyncRecursive = function(sourceDir, newDirLocation, opts) {
    opts = opts || {};

    try {
        if(fs.statSync(newDirLocation).isDirectory()) { 
            if(opts.forceDelete) {
            exports.rmdirSyncRecursive(newDirLocation);
            } else {
                return new Error('You are trying to delete a directory that already exists. Specify forceDelete in the opts argument to override this. Bailing~');
            }
        }
    } catch(e) { }

    //...
```

I'll submit a PR to wrench-js to honor the `preserveFiles` option and bypass the `rmdirSyncRecursive()` if it is set to `true`. This seems to be the best option.

What are your thoughts?

derekslife:
PR created for correcting this behavior within wrench:
https://github.com/ryanmcgrath/wrench-js/pull/100

jlukic:
Thanks @derekslife you present this all quite lucidly. 

I think we can point to your forked wrench copy as an NPM dependency, add your patch, then wait for a response to your PR. The project seems to be relatively idle, so I'm not counting on it being quick.

jlukic:
Merged into `next` in 1ca71b016d2d8bb014149443792193d003e5bdb7

jlukic:
I've tested install with changes and everything seems to be fine, files are preserved and new files are copied to existing directories.

derekslife:
Thanks. 

In the long term, due to wrench's stale project state, we should explore moving away from it all together. It would be more idiomatic to just use gulp's own vinyl-fs. However, we may have to submit a PR for this issue (https://github.com/wearefractal/vinyl-fs/issues/30) to get the overwrite behavior needed.

jlukic:
Agreed. Would prefer to use something native. Its always nice to replace a 10 liner with a 1 liner though when it makes sense.

derekslife:
PR Submitted: https://github.com/wearefractal/vinyl-fs/pull/59

PR #1891, opened 2015-02-28
Fix image deformation when using fluid or size classes

PLPeeters:
Currently, images which have height and width attributes set in the HTML don't resize correctly with the size and fluid classes (`ui small image`, `ui fluid image`, etc.). Those attributes are frequently used to avoid the content jumping around in case images take some time to load, which is why I figured this simple fix might be useful.

What happens in the actual version: the images get squeezed horizontally but not vertically, not preserving their aspect ratio.

What happens after my changes: the image gets resized both horizontally and vertically, keeping their aspect ratio intact.

### DIFF
--- a/src/definitions/elements/image.less
+++ b/src/definitions/elements/image.less
@@ -159,6 +159,7 @@ img.ui.bordered.image {
 .ui.fluid.image img {
   display: block;
   width: 100%;
+  height: auto;
 }
 
 
@@ -223,6 +224,7 @@ img.ui.bordered.image {
 .ui.mini.images svg,
 .ui.mini.image {
   width: @miniWidth;
+  height: auto;
   font-size: @mini;
 }
 .ui.tiny.images .image,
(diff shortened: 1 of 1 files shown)


jlukic:
Images [default to height auto](http://www.w3schools.com/cssref/pr_dim_height.asp), why would we need to specify?

PLPeeters:
I guess the height and width attributes override that behaviour, since it's causing problems for me without external CSS loaded.

jlukic:
Can you create a JSFiddle that demonstrates the issues you're seeing?

PLPeeters:
There you go: http://jsfiddle.net/cs93hfp3/

jlukic:
Are you working with backend tools that automatically add inline `width` and `height`? Its a bit quirky to specify a different height as an attribute than what you want it to appear as. 

Nevertheless there's no harm in adding. And it does solve this case.

PLPeeters:
Those are indeed added by my backend tools. Now that you point it out, it's indeed a bit quirky...

PR #1896, opened 2015-03-02
[Popup] Doesn't remove Window Events on Destroy

revov:
In an Angular JS application of mine I was suffering from a terrible memory leak. The extensive use of Semantic UI popups made me think they are the culprit. So I investigated a bit and found the cause. The **initialize** method of the Popup module attaches a listener on the **window.resize** event.but does not unbind it when its **destroy** behaviour is invoked. The proposed changes solve the memory leak.

### DIFF
--- a/src/definitions/modules/popup.js
+++ b/src/definitions/modules/popup.js
@@ -152,6 +152,9 @@ $.fn.popup = function(parameters) {
             .off(eventNamespace)
             .removeData(moduleNamespace)
           ;
+          $window
+              .off('resize' + eventNamespace, module.event.resize)
+          ;
         },
 
         event: {


jlukic:
We will have to solve this with `elementNamespace` (see [sidebar](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/sidebar.js#L130) for an example) since all modules bind to `window` with the same event namespace currently, causing the PR fix to remove events for other popups.

revov:
@jlukic This is why I pass the event handler to the 'off' method to make sure that only this event handler is removed and correspondent handlers for other popups are preserved (I made sure to test this before the pull request - other popups are intact).

However, you are correct that making the elementNamespace unique is a cleaner way to do this, since we can then unbind all events for this module in one go. Moreover, this would be more consistent as it was already done in other modules.

Do you want me to apply the *sidebar* approach from 9170fad1bd6e1e91c3a78aa391fe8516a32e8aee to my PR or you will take care of this?

jlukic:
You're right I didnt notice that in your commit, it does explicitly specify the event to unbind. 

I'm going to however swap over to a separate namespace to make it possible to remove the events explicitly without maintaining a reference to the event handler, which isn't possible outside the module itself.

PR #1978, opened 2015-03-18
Adding missing event namespace in form validation

hyperborea:
The namespace for e.g. the submit handler has been missed, which causes 'destroy' not to tear them down. 

I'm on Meteor which reuses DOM elements all the time, so I have to destroy / reinitialise my dynamic form all the time. This bug caused submit handlers to stack up.


jlukic:
Thanks for the fix! Merged this into `next`.

Be sure to merge future PR to that branch

Issue #1988, opened 2015-03-19
Required field indicators for grouped field labels should be on the group label


palmsey:
Demo: https://jsfiddle.net/3fd9wnkk/1/

The required field indicator (*) seems like it should go on the group label ("Would you like us to import your current settings?") rather than the individual radio button labels.

The markup in the fiddle is pulled from the docs, just added the "required" class to the grouped fields div.

gabormeszoly:
SUI fiddle here: http://jsfiddle.net/am0pgkqp/

palmsey:
Sorry, I forked that the first time, not sure what happened. 

Updated - http://jsfiddle.net/pktvcs1c/3/

jlukic:
I guess `require fields` should be each when `:not(.grouped)`

palmsey:
That's the solution I came up with too - I wasn't sure if I was missing some other way this was supposed to be set up though.

jlukic:
No, it was just an oversight. I'll fix

jlukic:
Added PR, closing issue

PR #1989, opened 2015-03-19
Moving required field indicator on grouped fields to group label

palmsey:
The required field style shouldn't apply to fields nested within grouped fields - #1988


jlukic:
Thanks for this

PR #2019, opened 2015-03-25
Disabled input field showing wrong cursor

nvnvenki:
Hi,

There was no indication to the user visually to say whether the field is disabled or not(Though slightly grayed out). So thought of adding disabled cursor so that user will immediately see the not-allowed cursor as user hovers over the input field.

Thanks,
Venki


jlukic:
Disabled styling is enough. I'm not a fan of modifying cursors.

PR #2034, opened 2015-03-28
Fix <select> inside form inline field

gutenye:
When you use an inline field, select should also be `inline` the same as input

Demonstrate at http://jsfiddle.net/tey5po7t/1


jlukic:
Thanks for the perfect PR format

PR #2047, opened 2015-04-01
use the optional keyword for imports

bassjobsen:
See http://lesscss.org/features/#import-options-optional, the optional keyword enables to drop the requirement for having many empty files.

### DIFF
--- a/src/theme.less
+++ b/src/theme.less
@@ -15,15 +15,15 @@
    Packaged Theme
 -------------------*/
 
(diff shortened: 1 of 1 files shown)


jlukic:
Hi, this is already in the [next release](https://github.com/Semantic-Org/Semantic-UI/commit/3df6034a0b524202346ec9c5fe58a37e458788a7).

I'm lagging about 200 commits behind in master.

PR #2062, opened 2015-04-04
Add onOpening callback to Accordion

clupprich:
I added a new callback to the Accordion module, named `onOpening`, which will be triggered before the Accordion menu is about to be opened.

### DIFF
--- a/src/definitions/modules/accordion.js
+++ b/src/definitions/modules/accordion.js
@@ -157,6 +157,7 @@ $.fn.accordion = function(parameters) {
           ;
           if(isUnopen) {
             module.debug('Opening accordion content', $activeTitle);
+            settings.onOpening.call(this);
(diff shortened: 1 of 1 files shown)


jlukic:
Added in `2`

zilev:
one question please, to which version will be available onOpening callback ??? I would like to use, but I cann't

jlukic:
The version coming out today.

zilev:
screenshot-3
I'm confused, in the official page http://semantic-ui.com/ the current version is 1.12.3 , and there isn't that callback added :(

jlukic:
Later today.. https://github.com/Semantic-Org/Semantic-UI/issues/2216#issuecomment-114606850

zilev:
ohh I see thank you so much :dancer:

Issue #1057, opened 2014-09-03
How toget 2/3 + 1/3 columns with grids ?


brunob:
Hi, i'm searching a way to get a grid with columns like that :  2/3 total width + 1/3 total width

This example show what i can do for now, but it's not exactly 2/3 vs 1/3 : http://jsfiddle.net/s1e48tn0/

Does anyone knows if it's possible ?

Thanks :)

MrFusion42:
Even when changing the `@columnCount`, one cannot have mixed grids (it would be hard to combine thirds with halves).

This codepen exemplifies some of the limitations: http://codepen.io/MrFusion42/pen/emajvo

Since this was a very important issue for my company, we worked on integrating grid mixins that act as generators, so that all of the needed CSS rules can be easily generated.

This is the same markup, but loading our [forked version](https://github.com/MrFusion42/Semantic-UI/tree/column-generators) of Semantic UI:  http://codepen.io/MrFusion42/pen/wBLKQj

I've submitted a pull request with the changes: [#2090](https://github.com/Semantic-Org/Semantic-UI/pull/2090).

MrFusion42:
Added pull request #2095 for the branch _"next"_.  (Sorry, didn't notice that pull requests were not supposed to be on master)

PR #2090, opened 2015-04-09
Add column generators, support for mixed grid spans

MrFusion42:
Regarding issue #1057, this is a proposed solution that generates all the permutations needed to mix any grid size _n_ with any column spanning (up to _n_).  (See codepen [with](http://codepen.io/MrFusion42/pen/wBLKQj) and [without](http://codepen.io/MrFusion42/pen/emajvo) this patch)

As a side-effect of this new feature, the generation of all other column-related rules is now automated (so for example setting `@columnCount: 18` generates the `.seventeen` and `.eighteen` classes and rules on forms and tables).

Generating rules for _n_ columns is as easy as:
```less
&{.forCols(); .-each(@number, @i) {
    .my.@{number}.column.grid > .foo.column {
        .-nColumn
    }
}}
```
This generates:
```css
.my.one.column.grid > .foo.column {
    width: 100%;
}
.my.two.column.grid > .foo.column {
    width: 50%;
}
.my.three.column.grid > .foo.column {
    width: 33.33333333%;
}

/* ... up to @columnCount, or pass a number to .forCols() */

.my.n-th.column.grid > .foo.column {
    width: 5%;
}
```
I documented the generators on `mixins.less`, but I guess that should go elsewhere.
 
If this PR is too disruptive, I can create a new one that just adds the needed rules to `grid.less` :smile:


pateketrueke:
:+1:

PR #2092, opened 2015-04-09
Implement filter function for transforming server responses

### LINKED ISSUE #2071: [search] Add support for transforming data from server response
Currently, if you are using the search module with a remote data source, it must return the correctly formatted JSON for the search module to be able to consume it. However, if the server does not return correctly formatted JSON, and if you can't modify the response the server sends server-side, then you can't use it with the search module.

There should be a way to transform the data received from the server response with a user-defined function client-side, the return value of which would then be passed into the search module (similar to [Typeahead.js's `filter`](https://github.com/twitter/typeahead.js/blob/master/doc/bloodhound.md#remote)). This would allow users to use the search module with any data source, provided that they transform it correctly.

mnquintana:
Resolves #2071

@jlukic Let me know what you think!

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -389,7 +389,6 @@ $.api = $.fn.api = function(parameters) {
               else {
                 settings.onSuccess.call(context, response, $module);
               }
-            }
             },
             error: function(xhr, status, httpMessage) {
               var
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -367,15 +367,22 @@ $.api = $.fn.api = function(parameters) {
             },
             done: function(response) {
               module.debug('API Response Received', response);
-              if(settings.dataType == 'json') {
-                if( $.isFunction(settings.successTest) ) {
-                  module.debug('Checking JSON returned success', settings.successTest, response);
-                  if( settings.successTest(response) ) {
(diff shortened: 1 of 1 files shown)


jlukic:
Will get back by early next week

jlukic:
Working on this currently

jlukic:
Added as `onResponse` this keeps callbacks names using similar naming like `onRequest`, `onSuccess` etc. 

I've modified the code to only add debug statements if a translated response is returned by the callback. 

```javascript
response = settings.filter(response);
```

is actually pass by reference, so it is the same as 
```javascript
settings.filter(response);
```

I've avoided other debug statements that use `response` from showing incorrect values by using a shallow copy `$.extend(true, {}, response)` for the returned value.

jlukic:
And I think we're good

mnquintana:
Awesome, thanks @jlukic! Any chance you could push a new build today?

jlukic:
This is all wrapped into the massive `2.0`. I'm only backporting bug fixes. You'll have to wait a couple weeks.

mnquintana:
@jlukic Sorry, I did mean on the beta channel - just a new build in the `dist/` folder.

jlukic:
rebuilt

jlukic:
http://beta.semantic-ui.com/behaviors/api.html#adjusting-server-responses

PR #2095, opened 2015-04-10
[Next] Add column generators, support for mixed grid spans

MrFusion42:
Regarding issue #1057, this is the same as pull request #2090 (see more details there), but for the brach _"next"_.

This enables grid column spans (`.number.wide.column`) to work as a _fraction multiplier_ (_i.e._ numerator) for any grid division (_i.e._ denominator), so that a `.two.wide.column` (2x) inside inside of a `.three.column.grid` (1/3) is `66.66%` wide (2/3).

**Codepen comparison:** [latest v2.0 build](http://codepen.io/MrFusion42/pen/wBLeQm) and [this PR fix](http://codepen.io/MrFusion42/pen/WbqOJy).


jlukic:
I'll get back on this. Have to have a think. The feature is necessary but my concern is with implementation. 

My intent was to avoid control structures whenever possible in css to ease in ports. Since the math  only has to be written out once, we could just include the repetitive output as well. Will get back next week.

MrFusion42:
I think that having the ability to change the `@columnCount` and have the rules adjust automatically is very valuable, so either there is no waste or there are no missing rules (right now there is no `.seventeen.column.grid` nor `.seventeen.wide.column`, even if `columnCount` is 17).

There's quite a lot of repetition too in color variations that could benefit from for-each loops.  Imagine if users could just extend/override a variable and add new colors everywhere.

SASS already has loops and control structures, so porting should be quite easy (better yet, they also get the generator goodness).  For other languages, a generated, repetitive version could be the one ported.

jlukic:
The aspiration is to have the ports automated as part of the build system which would require simple regex replace or pre existing conversion tools. This is already being done for NPM package conversions and meteor packages with gulp admin scripts.

I agree they provide much more flexibility and access to use cases unavailable to pure CSS, but they also add a layer of complexity. 

I think part of the issue is the default column count for SUI is too low, which was something I realized after 1.0 launched but didn't have any capacity to change without causing significant upgrade issues

MrFusion42:
I might be missing something, but it still seems to me that this would not be a problem:

* For pure-CSS distributions (such as the current ones for individual components, which, as far as I could see are the same ones for Meteor and NPM) the output will be CSS nonetheless, so whatever was on Less is not relevant anymore.  No problems here.
* For Less distributions, well, there is nothing to port, since it is already in Less.  No problems here.
* For SASS ports, that could be the most "problematic", but since it already has loops, a regex-only port could be achieved as long as some standard mixin syntax is followed.  I did not find any active SASS ports, however.
* JavaScript behaviors are not affected in any way by Less mixins, so no special considerations need to be taken.

I tested both `gulp "create components"` and `gulp "create distributions"`, which seem to be working fine (except for some apparently non-breaking errors displayed on the `next` branch as well).

**Alternatively**,
I could submit a new pull request without mixins (or nested rules for that matter) with just the additional grid/column-span permutation rules up to, perhaps, 20 columns (?).

The only difference would be that a `.two.wide.column` inside of a `.three.column.grid` would take the expected `66.66%` (2 / 3) instead of the current `12.5%` (2 / 16).

```html
<div class="ui three column grid">
  <div class="two wide column">I should be two thirds</div>
  <div class="column">I am one third</div>
</div>
```

Also interesting the current behavior of `.one.wide.column` vs `.column` inside of a n-column grid:

```html
<div class="ui three column grid">
  <div class="one wide column">Help! I'm one sixteenth!</div>
  <div class="column">I am one third</div>
</div>
```

jlukic:
I think that makes me feel ok about this is that a lot of people may want to roll their own `buttons` or `labels` etc, but very few people will ever need to roll their own grid.

pateketrueke:
> but very few people will ever need to roll their own grid since it is not really a design element but a structural element.

Yeah, of course it's structural, but many other frameworks already have those features.

I want to be more fluid while structuring stuff first since structure should not be a constraint for design, I guess.

Also I though that 2/3 (two thirds) are truly semantic, or isn't?

I switched to SUI for "semantic-ness goodies", so I hope to see this feature landed soon.

FYI; I'm using the @MrFusion42 fork without any issues.

MrFusion42:
Solved some merge conflicts in case you want to integrate this (yay!)

pateketrueke:
:+1:

MrFusion42:
The pull request has been updated to avoid merge conflicts (hint, hint)

pateketrueke:
We seriously need this asap! :+1:

gmulders:
Is there any chance that this will make it into SUI anytime soon?

kylevdr:
Hope this hasn't been abandoned. The current lack of ability to create 2/3 wide columns is pretty concerning.

john-osullivan:
@jlukic is this ever going to be merged?  I've been hacking around the inability to do layouts in thirds for a long time now.

taschetto:
This is so much needed...

Issue #2056, opened 2015-04-03
Running gulp in project folder, it looks for semantic.json in wrong path


miniskulljob:
Hi.

Let's assume we have Semantic-UI installed in `C:\projects\node_modules\semantic-ui` and we specified during install that our project folder is in `C:\projects\brand\web` and everything else default.
Then, we navigate to the `semantic` folder in our project and run `gulp build` like the installation suggests. It will throw an error like this:

```
Building Semantic
Cannot build files. Run "gulp install" to set-up Semantic
```

Trying to figure out the problem, I tried moving the `semantic.json` to the `C:\projects` folder, and it worked. So, my guess is that when we run `gulp`, it looks for its installation path, and then it starts from there, assuming its relative path is the installation path, so it looks for the `semantic.json` in the wrong place.
The only working solution I found, without moving the semantic.json file out of the project folder, is copying the `gulp` installation to a `node_modules` folder inside the project folder.

jlukic:
Hi, it'll continue climbing up looking for `semantic.json` and then use the paths relative to that location. 

So if you move the file up one level you'll need to adjust the paths in the file to include the additional directory.

miniskulljob:
Nope, that's the problem: It won't use the `semantic.json` path, it uses the `gulp` installation path.
```
C:\projects\node_modules\                   // Semantic-UI, gulp and other dependencies install path
C:\projects\brand\web\                         // Project path specified during install
C:\projects\brand\web\semantic.json   // JSON file generated during install
C:\projects\brand\web\semantic\          // Semantic-UI folder created during install
```
If we run `gulp build` in the `semantic` folder:
```
C:\projects\brand\web\semantic\gulp build

Building Semantic
Cannot build files. Run "gulp install" to set-up Semantic
```
It fails because it's taking `C:\projects\node_modules\` as the starting path and climbing up from there, and I know it is like that because if I put the `semantic.json` in `C:\projects\`, it works.

rudyryk:
Have the same issue, just performed fresh install via `npm install semantic-ui`. And I have got `gulp` installed globally with `npm install -g gulp`.

dattanchu:
This issue exists in semantic 2.0 release. The search is not done from the current project folder (where the gulpfile.js is) but from the node prefix folder.

jlukic:
Are you using NPM3? Peer dependencies should install gulp packages in your project folder.

dattanchu:
[...]

PR #2102, opened 2015-04-11
Update build.js

rudyryk:
I suggest more explicit error message on gulp build command fail, so developer will have an initial idea about what's wrong. Here's a related discussion and bug report: https://github.com/Semantic-Org/Semantic-UI/issues/2056


jlukic:
Thanks

PR #2156, opened 2015-04-22
Exclude list item link that are buttons from link color override.

ahtinurme:
This is purely my point of view, but list items that are link-type buttons text color should not be overrided. I've ran multiple times to this problem when I try to create a list of buttons and all the time the text on the link is in the wrong color.


jlukic:
I agree that's incorrect. I think it's probably safe to use the `:not(.ui)`. This is used elsewhere to avoid link stylings and is one of the nice side benefits of scoping elements to `ui`

Issue #2157, opened 2015-04-22
List item link button text with wrong color.


ahtinurme:
If you create a list with a item that's content is a button, the button text color is always overrided (see the last button in this jsfiddle: http://jsfiddle.net/1512gb5m/3/ )

Solution would be to exclude the .ui.button types from override.

jlukic:
Fixed in `2.0`

ahtinurme:
Will this also be backported to 1.x? Since it's also broken there.

jlukic:
No, everything is being lumped in together for a 2.0 release which will be in the next couple weeks.

ahtinurme:
Good :+1:

PR #2158, opened 2015-04-22
Exclude list item link that are buttons from link color override.

ahtinurme:
Fixes the #2157 issue by excluding .ui.button from a override.


jlukic:
Closed with #2157, passed along credit in release notes

PR #2168, opened 2015-04-24
Corrected order of operations to enable precision (rounding).

aaroncox:
I was attempting to put decimals in the progress bar (e.g. 9.4 of 10) and found that the precision rounding order of operations was out of order causing everything to round to 0 decimals. 

I corrected that order, however, many people may have gotten used to this behavior so the only additional change I considered was defaulting the precision to 0. 

I hope this helps!

### DIFF
--- a/src/definitions/modules/progress.js
+++ b/src/definitions/modules/progress.js
@@ -236,7 +236,7 @@ $.fn.progress = function(parameters) {
             if(settings.precision === 0) {
               return Math.round(displayPercent);
             }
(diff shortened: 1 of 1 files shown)


aaroncox:
I also added the precision rounding to the total value that can be used in the text label on the bars. So if the value is indeed `9.4 out of 10`, using `{value}` will put `9.4 of 10` instead of `9 of 10`

jlukic:
There was an issue with multiply total by zero when precision is not set. I've just resolved.

PR #2183, opened 2015-04-27
"wrench" depencency as tarball

zenflow:
Please use https/tarball for wrench dependancy instead of git protocol, as it is more portable and doesn't require git to be installed.


jlukic:
Yep, much better will adjust.

PR #2209, opened 2015-05-05
Ordered list does not invert number color.

pcj:
See http://jsfiddle.net/ka98j2pv/2/
This commit introduces @orderedInvertedCountColor for this purpose.


pcj:
F me.  Bad commit.

PR #2210, opened 2015-05-05
Make ordered list numbers in an inverted context be visible

pcj:
See http://jsfiddle.net/ka98j2pv/2/
This commit introduces @orderedInvertedCountColor for this purpose.
@jlukic If prefer to make changes on your branch and close this PR, no biggie.


jlukic:
Wonderful, good catch. Added to `next`

PR #2225, opened 2015-05-07
The wrong this

pencilcheck:
The this in the call to refresh() from MutationObserver is the module itself not the DOM jquery object. Changing to element fixed it.


jlukic:
Yeah this was just incorrect, thanks

jlukic:
This appears to have already been fixed in `next`, but not backported.

PR #2228, opened 2015-05-08
Html5 video

challet:
Adding an ``htmlvideo``module which is essentially a Javascript behavior : binding control elements to a video element : 
* playpause button
* forward/rewind buttons (some browsers don't fully support it, see in-code comments)
* display current/remaining time
* time range
* (wip for more)

Adding an example page with the previously described features and impemented through SUI / HTML elements :
* *.ui.button*
* *.ui.statistic*
* *input type="range"* 

Sorry that the *dist* folder has been commited along with it.


jlukic:
Hi, there's currently 131 files changed. Could you make sure this matches current `next` branch with only modified `src/` files. 

You can try
```
git checkout dist/ 
```

challet:
see #2230 for cleaner version

jlukic:
You can also feel free to use the name `video`. All modules are single most common english words. 

I'm currently working on retooling the embed component in branch [`embed`](https://github.com/Semantic-Org/Semantic-UI/blob/embed/src/definitions/modules/embed.js)

It will support any third party content that can be hidden behind a placeholder image, this is to avoid including external APIs until a user confirms he wants to activate the content.

Another thing I've noticed, be sure to organize method calls in verb -> object order, for example 

```javascript
time: {

   current: function() {

  }
}
```

Works better for external APIs if you use

```javascript
current: {

   time: function() {
     // get time
  }
}
```

PR #2230, opened 2015-05-08
Html5 video

challet:
cleaned PR, though I find myself unable to correctly launch a build with the new module included, i don't know exactly why.

Adding a video module which is essentially a Javascript behavior : binding control elements to a video element :
- play/pause switch button
- forward/rewind push buttons
- display current/remaining time
- time range "slider"
- volume up/down/progress/mute
- (wip for more)

Adding an example page with the previously described features and impemented through various SUI / HTML elements : http://html5video.challet.eu/examples/video-player.html


jlukic:
Go ahead and take over the `video` namespace so you dont have to deal with adjusting build tools.

You have to add to task default component list, this is used for interactive installer when selecting custom builds, and for setting components array when none is specified
https://github.com/Semantic-Org/Semantic-UI/blob/master/tasks/config/defaults.js#L44

challet:
About the naming from the previous PR, how could I name the method which links an html element to display and autoupdate the time in ? "add display current time " ?
then "add control volume up" etc. ?

jlukic:
You need to use the `selector` settings object for specifying internal selectors of your component, i.e rewind, fast foward, etc.

You should also move all event binding to a `bind: { events: function() {}`, see the annotated source code for explanations

src/definitions/modules/video.js
```diff
@@ -85,20 +93,218 @@ $.fn.video = function(parameters) {
...
+                $playpause.addClass('active')
+              }).on('pause' + eventNamespace, function() {
+                $playpause.removeClass('active')
+              });
+          },
+          
+          // see https://developer.mozilla.org/fr/docs/Web/API/HTMLMediaElement#playbackRate
+          // mode from 'switch', 'push'
+          playbackrate: function($playbackrate, rate, mode) {
+            $playbackrate = $($playbackrate);
```
jlukic: Don't require the user to specify all controls themselves. Use names specified in `selector.thing`, i.e. `selector.playrate`, then only query for this once in top of module cache (see annotated source).

src/definitions/modules/video.js
```diff
@@ -85,20 +93,218 @@ $.fn.video = function(parameters) {
...
+            // from UI to video
+            switch(mode) {
+              case 'push': default:
+                $playbackrate.on('mousedown' + eventNamespace, function() {
+                  element.playbackRate = rate;
+                });
+                $playbackrate.on('mouseup' + eventNamespace + ' mouseleave' + eventNamespace, function() {
```
jlukic: Don't use any anonymous functions, if it requires the `event` object use the `event: { }` naming scheme, see other modules, otherwise it should trigger an internal action, for example, clicking on playback rate should call something like `module.change.playbackRate`

challet: About this example with ``module.change.playbackRate``: how is the best to configure the 2 buttons (forward and backward) that would both call the same function, but with a different rate (say 6 and -4) ? Should this setting be inside the module or in a ``data-playback-rate``attribute for each of them ?

challet:
Thanks for your advices, that also made the events flow and API a lot more clearer !

challet:
Hi jlukic, I've added a few things, and I'm now wondering how to actually display the [TimeRanges data](https://developer.mozilla.org/en-US/docs/Web/API/TimeRanges) since they can split into several sub ranges. It would apply to buffered, seekable and played properties of the [video element](https://developer.mozilla.org/fr/docs/Web/API/HTMLMediaElement#Properties).
I would also be grateful if you can have a code review on what's been commited.

jlukic:
I'll take a more in depth look at this later in the week.

challet:
I'm thinking about to add an "video state api" to the module by introducing functions to the "get ..." range, used for both internal event handler and external calls.  
That would prevent to read the dom on each request, and would make it more versatile if the main video element is different, say a proprietary flash video element with it's own dom api. Any thought/recommendation about this ?

challet:
Hi @jlukic, here are what I can think about as additions : 
* Video track url picker/editor element
* Timeranges display : SVG would be great, but isn't it far from the core SUI techs ?
* Subtitles : I have to think about a code refactoring, but is it useful (as being part of the module) ?
* More various exemples
* Unit tests
* Doc

jlukic:
Hi I apologize for the delay. The launch deadline of June 1st prohibits my attention from any non critical updates. I will be able to continue looking at this in June for `2.x`

challet:
No problem, I keep on looking for additions

src/definitions/modules/video.js
```diff
@@ -68,12 +131,11 @@ $.fn.video = function(parameters) {
...
         initialize: function() {
           module.debug('Initializing video');
-          module.create();
-          $module
-            .on('click' + eventNamespace, selector.placeholder, module.play)
-            .on('click' + eventNamespace, selector.playButton, module.play)
-          ;
           module.instantiate();
+          module.bind.pushes();
+          module.bind.popups();
```
challet: is it a right way to call sub modules ?

jlukic:
I'm not sure what you're doing, but everything should be just one module..

challet:
I know sorry I made it out of asking : this `push-button` branch and module is an abstracted behavior for *send periodic clicks until the butoon remains pushed*.
It is not absolutely required by `video`, but the example works a lot better with it : it is used both for loop seeking (forward and rewind) and for volume up/down. And that ended up being a module. 

It might be better somewhere else, but I don't really know where. The code of this "push" behavior is three function and we'll be able to move it back to "video" if needed. It also might be the subjet of a new PR, but since one used the other, I dond't know how to split it in two PRs.

jlukic:
Some notes. Don't have time to do a formal review

* Don't add any other modules or namespaces. I don't think setting an interval on click warrants a new component
* Use `settings.metadata` for html5 data names
* Don't add any free floating utility functions
* Try running JS linting like JSHint, there appears to be some mixed white space and tabs and a couple missing semicolons.
* Make sure log statements are meaningful, no reason to log anything in an `is` state unless you are displaying some related debug data.
* No need for helper functions for attaching event namespace
* Check line breaks on if/else statements and case statements

challet:
Thanks for your notes, I'll be available for a more in-depth review when you have time.
* About the new module : I've used it for one thing, it is to store the variables, timers, states etc. (well, a module) in a html element. I'll try to change it back and to store them from the `video` module 
* utility functions : should i add them as being part of the module API ? (i'll remove the attaching events one)

I'm working on the other issues.

challet:
Hi @jlukic, what about to move this *push* behavior in the ``button`` namespace ? If that's a final no, I'm moving it back to ``video``.
Any more comment on the code, exemple interface, api etc. is welcome.

src/definitions/modules/video.js
```diff
@@ -83,196 +104,448 @@ $.fn.video = function(parameters) {
...
-              }
-              else if(source == 'youtube') {
-                html = ''
-                  + '<iframe src="//www.youtube.com/embed/' + id + '?=' + module.generate.url(source) + '"'
-                  + ' width="100%" height="100%"'
-                  + ' frameborder="0" webkitAllowFullScreen mozallowfullscreen allowFullScreen></iframe>'
-                ;
-              }
+        
+        // the functions in this range will change the video element state, with no relevant return value
```
challet: or should they return a true/false bit on the success of the operation ? It would be implemented through try/catch blocks for each dom operation I guess.

challet:
Remaining to be done : 
* autoplay/loop/preload checkboxes reflecting prop values
* fullscreen
* handle ```<track>```elements, not sure what to do here, see https://developer.mozilla.org/fr/docs/Web/HTML/Element/track
* more ?

brackmayhall:
Where do we stand on the html5 video? I would love to use semantic-ui but I need html5 video support for my project.

PR #2260, opened 2015-05-15
fixing dropdown showing when page is re-focused

trevorharwell:
It is very awkward to have the dropdown re-appear after you select an option, change to another page, and than tab back to the browser.  The fix is very simple and seems to be more natural behavior.


jlukic:
Hi, everything checks out, and I've added this to `next` release branch with credit in release notes. 

Check readme notes on PR for next time. No biggie though.

PR #2308, opened 2015-05-28
Refresh popup content on 'refresh'

ascariandrea:
Hi, first of all I want to congrats for your project and maybe I'm wrong of implementing this feature because I've missed something.

I create this pr 'cause I need to refresh the content in the inline popup when i change the trigger `data-html` attribute.
So I changed the jquery `data()` method with the `attr()` cause the first one uses cache and don't refresh the popup content.
A more detailed answers can be found [here](http://stackoverflow.com/a/8708345).

Hope it will be helpful,

### DIFF
--- a/src/definitions/modules/popup.js
+++ b/src/definitions/modules/popup.js
@@ -14,6 +14,7 @@
 "use strict";
 
 $.fn.popup = function(parameters) {
+
   var
     $allModules    = $(this),
     $document      = $(document),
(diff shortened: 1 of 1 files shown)


jlukic:
Can you create a JSFiddle to show how the popup content is not adjusting correctly when metadata attributes are changed?

ascariandrea:
Maybe I should say that I'm using the popup modules in a React component.
Anyway this is the [jsfiddle](https://jsfiddle.net/0pkvvnj0/2/).

jlukic:
Thanks for your pull request. I appreciate you taking the time to solve your issue for others using the library. 

In the end I've elected to use a slightly simpler solution, clearing cached metadata before get each value. This will be included with `2.0` next week.

jlukic:
Example
http://jsfiddle.net/zye3a87j/

PR #2331, opened 2015-05-31
fix font format

Jayin:


jlukic:
Youch, not sure how that happened

Issue #662, opened 2014-02-24
OS X Firefox renders icons too bold.


eiriarte:
Hi,

There's some issue with OS X Firefox which makes icons look uglily thick, specially inverted icons:

http://jsfiddle.net/eiriarte/7a58w/

It's seems to be solved using `-moz-osx-font-smoothing: grayscale;` (in my Macbook works well, see the screenshot).

More info here:

http://maximilianhoffmann.com/posts/better-font-rendering-on-osx

PR #2347, opened 2015-06-02
Fixes icon rendering problem in OSX + webkit

davialexandre:
The issue is pretty much the same as #662, but now in webkit. Using -webkit-font-smoothing: antialiased solves the problem.

I made a jsfiddle that demonstrate it: http://jsfiddle.net/davialexandre/trx0bpz8/. Here is an image of how the icons are rendered (I am using a dark background to make it easier to notice the difference). The first row is using -webkit-font-smoothing and the second one is how the icons are currently rendered:

captura de tela 2015-06-02 as 14 47 41

### DIFF
--- a/src/definitions/elements/icon.less
+++ b/src/definitions/elements/icon.less
@@ -53,6 +53,7 @@ i.icon {
   speak: none;
   font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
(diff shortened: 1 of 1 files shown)


kalebdf:
:+1: To pull this in.

We did this locally. Improved customer conversion 200%

PR #2369, opened 2015-06-06
Update login.html

lemartialou:
Small fixes: title and logo image path.


jlukic:
Thanks for the catches as always

lemartialou:
It's the least I can do ^^

PR #2396, opened 2015-06-11
Allow .dropdown() to respect disabled <select> <option>s when dynamically creating menus from select inputs.

malacalypse:
Currently, .dropdown() does not respect the "disabled" attribute on the <option> tag when creating dropdown menus from <select> elements. This patch fixes that, ensures that click events do not get handled (both in JS and in CSS), and appropriately colours the inputs based on @itemColour so that individual opacity changes do not cause background bleed-through.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -533,14 +533,14 @@ $.fn.dropdown = function(parameters) {
           },
           keydown: function(event) {
             var
-              $currentlySelected = $item.not(className.filtered).filter('.' + className.selected).eq(0),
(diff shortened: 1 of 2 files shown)


malacalypse:
If I've applied this to the wrong branch, please forgive me! I can re-apply to a different branch if you prefer.

malacalypse:
Further suggestions I could incorporate: 
Use variables for disabled lightness, disabled colour, and disabled background.

jlukic:
Hi, master is about 4 months behind `next`. 

`disabled` is respected, although it is not generated currently from `<select>` `disabled` property. This part of PR I'll look into.

Issue #2413, opened 2015-06-16
Output css and js to separate dist directories


kevindice:
_About to submit a PR for this._

I think it would be helpful to most people to have the ability to output css and js to separate directories.

kevindice:
Don't pull this yet.  After a whole project regex search, I realized I forgot to make necessary changes to  tasks/watch.js

jlukic:
#2221  Please be sure to check any PR against `next` and remember legacy support for existing `semantic.json` is essential

kbajpai:
Did this issue got fixed?

fuzenco:
Looking through the 2.2 documentation and Learn Semantic, I don’t see any way of defining the paths. Has this not been implemented yet?

kevindice:
No, not yet.

PR #2414, opened 2015-06-16
Output css and js into seperate dist directories

kevindice:
#2413


jlukic:
Hi, just need to make sure this works with legacy `semantic.json` file which do not contain an object for `css` and `js`.

Also a bit confused on use of `path.posix.join`

kevindice:
Removed the .posix part and added a mechanism for gracefully continuing with a deprecation notice.

jlukic:
Thanks, give me some time to review.

kevindice:
Would it be better if this did not append /css and /js directories to the output paths of existing installations?

jlukic:
Yes, that would be better. 


The install script should probably go 

* Different paths for js and css? (Yes/no)

If yes
* What path component js? 
* What path for component css?
* What path for concat js?
* What path for concat css?

If no, current questions.

PR #2416, opened 2015-06-17
Allow link underline and hover link underline to differ

morty:
I'm working on a site where I need links to be underlined when hovered, but otherwise not.

This patch allows you to add @linkHoverUnderline to your site.variables to make this possible.


jlukic:
Simple, obvious, thanks

PR #2477, opened 2015-06-29
Fix naming conflict in icon element

mxth:
It conflicts with global `@fontName` variable. This makes the font icon lost when overriding `@fontName` in the global.

### DIFF
--- a/src/themes/default/elements/icon.variables
+++ b/src/themes/default/elements/icon.variables
@@ -6,14 +6,14 @@
    Icon Variables
 --------------------*/
 
-@fontName: 'icons';
-@fallbackSRC: url("@{fontPath}/@{fontName}.eot");
+@iconFontName: 'icons';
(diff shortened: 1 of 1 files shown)


jlukic:
Component variables extend globals. I don't understand how changing the global font name should affect icons.

mxth:
I'm not quite sure. If so, the problem should be elsewhere. You can reproduce it by simply putting `@fontName          : 'Lato';` in `src/site/globals/site.variables`. Looks like it will override the `@fontName` of icons.

jlukic:
I've thought this over a bit tonight. I'm fairly certain the order of inheritance needs to be swapped. 

The previous order was:
* Default site variables
* Default component variables
* Theme site variables
* Theme component variables
* Site site variables
* Site component variables

The new order is:
* Default site variables
* Theme site variables
* Site site variables
* Default component variables
* Theme component variables
* Site component variables

In any case I could come up with component variables should always take precedence over site.variables. I've tested with your case and it fixes things.

PR #2492, opened 2015-07-01
Fixes "SyntaxError: unreachable code after return statement"

rodrigoddalmeida:
This error was being triggered in Firefox.


jlukic:
Thanks!

PR #2502, opened 2015-07-02
[Form] Add background-color transition to form inputs

vennamo:
This is present in elements/input.variables so it should be here too.


jlukic:
I'm fairly certain this causes issues with autocomplete styles

jlukic:
I can confirm this is omitted from `form.variables` because it causes autofill form styles to flash from the default browser yellow to the semantic UI yellow.

PR #2566, opened 2015-07-09
Fix icon not shrinking when circular icon used

vennamo:


jlukic:
Sorry ended up solving this in an issue before evaluting this PR. This is available in stable.

PR #2573, opened 2015-07-09
Properly set the dropdown value when dealing with remote loaded choices

davialexandre:
This problem happens only when we have a dropdown created from a select
that loads remote content. In this scenario, we don't know in advance
which choices will be available, so we start with a select without any
options. When an choice is selected, module.set.value will try to set
the select value, but it will fail, since there's no options.

To fix this, module.set.value will add a new option to the select, before
setting the value. This will be done only if the dropdown input is a select
and if we are dealing with remote loaded content.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2090,6 +2090,12 @@ $.fn.dropdown = function(parameters) {
                   return;
                 }
               }
+
+              if($input.is('select') && settings.apiSettings) {
+                module.debug('Adding an option to the select before setting the value', value);
+                module.add.optionValue(value);
+              }
+
               module.debug('Updating input value', value, currentValue);
               $input
                 .val(value)


jlukic:
On review it seems it should do if either `allowAdditions` or `apiSettings` is used.

davialexandre:
It makes sense. The only problem I see is that e61b39c16b73160d03165a185433b074a4be693f ended up duplicating code. Both module.set.value and module.add.value now have this:

```javascript
if($input.is('select')) {
  if(settings.allowAdditions || settings.apiSettings) {
    module.debug('Adding value to select', addedValue, newValue, $input);
    module.add.optionValue(addedValue);
    }
}
```
What about moving this check to module.set.select, inside the $selectedItem.each callback? I think it is a good way to remove this duplication.

jlukic:
* Added `can.extendSelect` 
* Dropdown now removes option values if a user/api selection is removed
* Added error conditions for multiple `<select>` initialized with multiple prop (caused me lots of headaches)

PR #2581, opened 2015-07-10
Fixes #2572

### LINKED ISSUE #2572: Toggle style checkboxes initialized with JS don't work
https://jsfiddle.net/pwkayddm/2/

.ui.toggle.checkboxes don't work when initialized with .checkbox() anymore. 

I've included both a rails-style checkbox (two inputs - one hidden, one a checkbox) and a traditional single-element checkbox, both written in the traditional inline style used by simple_form among many other libraries.

These used to work just fine in Semantic-UI 1.x , but now their toggle behaviour isn't correct - it appears to be double triggering. I thought it was simply because the onClick is bound to the label, which usually causes a double call on the input itself, but fixing that didn't entirely solve the problem.

malacalypse:
Since the input state is toggled by set.checked() the default behaviour of this click does not need to happen. 

This fixes the multiple trigger bug with labels and input event bubbling, while permitting input state change and bindings to still work.

### DIFF
--- a/src/definitions/modules/checkbox.js
+++ b/src/definitions/modules/checkbox.js
@@ -177,6 +177,7 @@ $.fn.checkbox = function(parameters) {
               return;
             }
             module.toggle();
+            event.preventDefault();
           },
           keydown: function(event) {
             var


jlukic:
Thanks for fixing this bug.  In future, please submit PR against `next` branch this way I can merge your commits and you can get credit. (Check readme for details).

malacalypse:
Roger that. I'll get the PR right next time. Thank you for the credit!

PR #2596, opened 2015-07-11
Update form.less

avalanche1:
form error dropdown should have highest priority, otherwise some of my styles override error styles. like this:
image
should be:
image


jlukic:
You should be modifying colors using variables.

avalanche1:
Well, unfortunately I cannot build SUI properly for some obscure errors, thus I have to override using css...
But anyways, doesn't error styling should have highest priority?

jlukic:
As long as inheritance works with in-library changes, I don't think it's necessary to adjust specificity.

Issue #2441, opened 2015-06-22
[Image] Add Greyscale


geordee:
How about adding a grayscale class for images?

    filter: url("data:image/svg+xml;utf8,<svg xmlns=\'http://www.w3.org/2000/svg\'><filter id=\'grayscale\'><feColorMatrix type=\'matrix\' values=\'0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0 0 0 1 0\'/></filter></svg>#grayscale"); /* Firefox 10+, Firefox on Android */
    filter: grayscale(100%);
    filter: gray; /* IE6-9 */
    -webkit-filter: grayscale(100%); /* Chrome 19+, Safari 6+, Safari 6+ iOS */

PR #2616, opened 2015-07-13
Added filters for image

Illyism:
grayscale, invert and blur


madfriend:
CSS `filter` does not work in IE (any IE) (http://caniuse.com/#search=filter)

jlukic:
Although these are neat and easy to use features, I think this is generally a bad thing to do in the browser for performance.

Illyism:
@jlukic Alright, no worries, you should close #2441 as well then.

PR #2645, opened 2015-07-15
[Search] Allow custom display fields

anibalmf1:
This change allow the developter to choose the field to display in the results. So we don't have to change our json's object field names to the fixed ones (title, price and description).
usage:

$('div[name="searchDiv"]').search({
            source: [{id:0, myfield: 'some text', extrafield: 'ST'}, {id:1, myfield: 'another text', extrafield: 'AT'}],
            searchFields: [
                myfield
            ],
            displayField: extrafield
        });

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -824,7 +824,7 @@ $.fn.search = function(parameters) {
               }
             }
             if($.isFunction(template)) {
-              html = template(response);
+              html = template(response, settings);
             }
             else {
               module.error(error.noTemplate, false);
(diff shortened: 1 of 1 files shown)


jlukic:
Thanks I'll review this with `2.1`

jlukic:
I've decided to implement this in a slightly different way, but the gist is the same.

jlukic:
All server fields are now a setting and can be modified at will. 

I've added examples to the docs as well on how to do this. 

Here's a quick fiddle to demo.
http://jsfiddle.net/npyfwkLa/

anibalmf1:
Nice work man... Congrats.

jlukic:
Added to dropdown with `remoteData` as well

Issue #2653, opened 2015-07-16
Error: Task watchui can't support dependencies that is not an array of strings


TehCrucible:
Another issue with importing tasks from main repo into my own gulp file.  Note that I've changed my task name to `watchui` rather than `watch ui`. Throws this exception:
```
C:\Users\jamie_000\OneDrive\Crucible\Cairns Key\website\cairnskeyTheme>gulp watchui
C:\Users\jamie_000\OneDrive\Crucible\Cairns Key\website\cairnskeyTheme\node_modules\gulp\node_modules\orchestrator\index.js:47
                        throw new Error('Task '+name+' can\'t support dependencies that is not an a
                              ^
Error: Task watchui can't support dependencies that is not an array of strings
    at Gulp.Orchestrator.add (C:\Users\jamie_000\OneDrive\Crucible\Cairns Key\website\cairnskeyTheme\node_modules\gulp\node_modules\orchestrator\index.js:47:10)
    at Object.<anonymous> (C:\Users\jamie_000\OneDrive\Crucible\Cairns Key\website\cairnskeyTheme\gulpfile.js:32:6)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Liftoff.handleArguments (C:\Users\jamie_000\AppData\Roaming\npm\node_modules\gulp\bin\gulp.js:116:3)
    at Liftoff.<anonymous> (C:\Users\jamie_000\AppData\Roaming\npm\node_modules\gulp\node_modules\liftoff\index.js:192:16)
```
Seems to be fixed by removing the text descriptions of the tasks from the suggested code like so:
```
// import task with a custom task name
gulp.task('watch ui', watch));
gulp.task('build ui', build));
```

jlukic:
Yeah, i've revised the examples. The second parameter is for people using [gulp help](https://www.npmjs.com/package/gulp-help)

fholzer:
Seems like there's still an issue with this...
1. gulp-help is still required, because your commit added `gulp.task` with the gulp-help parameter to `tasks/collections/rtl.js`
2. The documentation for this needs a minor fix too.

PR #2668, opened 2015-07-19
fix build/watch for people importing tasks into their own gulpfile

fholzer:
See Semantic-Org/Semantic-UI#2653

### DIFF
--- a/tasks/build.js
+++ b/tasks/build.js
@@ -3,7 +3,7 @@
 *******************************/
 
 var
-  gulp         = require('gulp'),
+  gulp         = require('gulp-help')(require('gulp')),
 
   // config
   config       = require('./config/user'),
--- a/tasks/watch.js
+++ b/tasks/watch.js
@@ -3,7 +3,7 @@
 *******************************/
 
 var
-  gulp         = require('gulp'),
+  gulp         = require('gulp-help')(require('gulp')),
 
   // node dependencies
   console      = require('better-console'),


fholzer:
This way, the gulp instance that is handed down to `rtl.js` supports gulp-help, as it adds new tasks to gulp.

fholzer:
In case you're worried about idempotence of gulp-help, I tested it. You could do this
````javascript
var gulp = require('gulp-help')(require('gulp-help')(require('gulp-help')(require('gulp'))));

gulp.task('default', 'this is some default task', false, function(done) {
        console.log('some task output');
        done();
});
````
and still get the expected output
````
storm@linux-krxl:~/git/semanti-ui/patch/test> gulp 
[09:22:29] Using gulpfile ~/git/semanti-ui/patch/test/gulpfile.js
[09:22:29] Starting 'default'...
some task output
[09:22:29] Finished 'default' after 107 μs
storm@linux-krxl:~/git/semanti-ui/patch/test> gulp help
[09:26:40] Using gulpfile ~/git/semanti-ui/patch/test/gulpfile.js
[09:26:40] Starting 'help'...

Usage
  gulp [TASK] [OPTIONS...]

Available tasks
  default  this is some default task
  help     Display this help text.

[09:26:40] Finished 'help' after 978 μs
storm@linux-krxl:~/git/semanti-ui/patch/test>
````

fholzer:
Sorry, didn't notice your comment on my other PR. Will close this now.

jlukic:
How are you importing tasks, with two parameters or three? 

I think it might be a bit idiosyncratic to require other people who import the tasks to use `gulp-help`

fholzer:
The alternative is to remove the gulp-help argument from gulp.task from all files that get sourced by `tasks/build.js` and `tasks/watch.js`. That would be nearly all tasks as can be seen when running `gulp help` in my [fixed version](https://github.com/fholzer/Semantic-UI-Issue-Showcase/tree/2653-fixed):
````
gulp help
[12:31:11] Using gulpfile ~\Documents\git\semantic-ui-tests\gulpfile.js
[12:31:11] Starting 'help'...

Usage
  gulp [TASK] [OPTIONS...]

Available tasks
  build
  build-assets                   Copies all assets from source
  build-css                      Builds all css from source
  build-javascript               Builds all javascript from source
  build-rtl                      Watch files as RTL
  default
  help                           Display this help text.
  package compressed css
  package compressed docs css
  package compressed docs js
  package compressed js
  package compressed rtl css
  package uncompressed css
  package uncompressed docs css
  package uncompressed docs js
  package uncompressed js
  package uncompressed rtl css
  watch
  watch-rtl                      Build all files as RTL

[12:31:11] Finished 'help' after 2.26 ms
````
As these come from files that are also sourced when running gulp from the semantic directory, removing the gulp-help argument from them would break running `gulp help` in the `semantic/` directory.
While keeping gulp-help around, *and* fixing the issue for users not directly using gulp-help in their `gulpfile.js` is just a matter of editing two files, and `gulp help` would still show all tasks when run from the `semantic/` directory.

The nice thing about my PR is that users aren't forced to use gulp-help in their gulp file, if they don't want to. Though even when they to, it will still work.

Negative side-effect: When users use gulp-help in their project, they also see all the semantic targets.
Though in order to prevent that, users shouldn't `require()` the semantic task files at the top of their `gulpfile.js`, but only in the task function that actually runs those tasks, [like so](https://github.com/fholzer/Semantic-UI-Issue-Showcase/commit/a8356f00ef3ee10f0c316a920ac19e61ab1f88ca). (Of course this also works file for people that don't use gulp-help.)

jlukic:
I'll be evaluating this shortly.. hold tight

jlukic:
I've tested this and it works as you say, using either gulp help or no gulp help in the imported `gulpfile.js` now works fine.

jlukic:
I've created an example repository which I will include in docs to help people understand what to do.
https://github.com/Semantic-Org/Example-External-Gulpfile

PR #2734, opened 2015-07-27
Add missing background-color variable for tabular menu

frontdevde:
The variable @tabularBackgroundColor was referenced in menu.variables but not actually used in menu.less

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -600,7 +600,7 @@ Floated Menu / Item
 ---------------*/
 
 .ui.tabular.menu {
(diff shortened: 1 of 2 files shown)


jlukic:
Hey nice catch. I'm assuming the variable should also be set to transparent? https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/collections/menu.variables#L303

frontdevde:
@jlukic Sure, amended the new default value for@tabularBackgroundColor.

jlukic:
Thanks looks good!

PR #2749, opened 2015-07-30
Add href="#" and replace div button by a tag

Tonours:
Add href="#" on a tag and replace 
```
<div class="ui huge primary button">
with 
<a class="ui huge primary button">
```


jlukic:
I think it's fine to omit the href="#" the links will still act the same in template. 

The one `<div>` button I agree should change though

Issue #2754, opened 2015-07-31
[Dropdown] Number valued items can not be removed after other remote search


dgurkaynak:
I'm trying to use dropdown module with multiple selection and remote content. And I came across with a bug. If item value is a number, after another search (when new items arrived), previously added items can not be removed. 

Fiddle: https://jsfiddle.net/juswr4o0/

To reproduce:
- Search and add an item.
- Clear query, search and add another item.
- Then try to remove items by pressing backspace or clicking `x` icon on the labels.

Demo:
![demo](http://i.imgur.com/xRxTve8.gif)

After a little bit digging, I realised that there is no problem with string values. I guess the problem is `jQuery.data()` is returning number value and semantic's dropdown module searchs that value in array with string values.

My workaround is treating all the values as string.

Btw, testing with v2.0.7.

jlukic:
Re-implemented using very similar code, credit given. Fixed in `next`

PR #2755, opened 2015-07-31
[Dropdown] Fix removing number valued items from dropdown (#2754)

dgurkaynak:
#2754


jlukic:
Hi, this should be merged with `src/` paths. I've re-implemented and passed on credit in release notes.

PR #2769, opened 2015-08-03
[Dropdown] Allow `onHide` callbacks to prevent Hiding

radiohead:
I found out that there is no way to prevent dropdown from being hidden in specific cases, so I thought this might be quite handy.
In my scenario, I use an input field with attached `jquery-datepicker` and whenever I want to jump between month in it, the dropdown hides.

Therefore I propose to add this small feature.


jlukic:
I agree, if you remove var declaration and put `settings.onHide` in the conditional, i'll go ahead and merge.

radiohead:
@jlukic Done, thanks for the note.

Issue #1549, opened 2015-01-02
Flags: England is not an alias for United Kingdom (Great Britain)


garethjmsaunders:
The United Kingdom of Great Britain and Northern Ireland is a sovereign state comprising four countries (in order of geographical size): England, Scotland, Wales and Northern Ireland.

The flag of the United Kingdom commonly known as the Union Jack or Union Flag comprises the red cross of St George (patron saint of England), edged in white, superimposed with the red cross of St Patrick (patron saint of Ireland) and the white saltire of St Andrew (patron saint of Scotland) on a blue background.

This is the flag (the Union Jack) that you have in your code:

```
<i class="gb flag"></i>
<i class="england flag"></i>
<i class="united kingdom flag"></i>
```

My issue is that **the flag of the United Kingdom is not the flag of England** and therefore `england` should be removed as an alias for the UK flag. Using `england` is a class name for this flag is at best un-semantic and at worst just plain wrong, and many people in Scotland, Wales, Northern Ireland and indeed England itself would find it offensive.

The flag of England is a red cross on a white background. See [Flag of England on Wikipedia](https://en.wikipedia.org/wiki/Flag_of_England).

A better, and more semantic class name would be: `uk` and/or `great britain`.

jlukic:
Yep, mistake, offensive.

PR #2770, opened 2015-08-03
[Flag] Add England Alias

eduardom:
Hi, my first (and tiny) pull request here. Thanks for the great work on Semantic-UI. Issue #1549 removed the alias uk-england but england itself was never added back.


jlukic:
Thanks for the catch.

MrFusion42:
Shouldn't the change be in `themes/default/elements/flag.overrides` instead of `dist/components/flag.css`?

jlukic:
Yes.. didnt notice path. I'll have to adjust

PR #2772, opened 2015-08-03
[Popup] Look for inline popup in all siblings of target, not just immediate sibling

malacalypse:
Specifying inline:true to popups should find .ui.popup in all siblings, not just the immediately following one (sometimes it isn't possible to guarantee that the popup is written immediately next, esp. with ajax content and complex sites).


jlukic:
Makes sense.

PR #2795, opened 2015-08-06
[Dropdown] Ignore Input change event on load if specified in settings

lauri-elevant:
I am trying to listen to the "change" events on all selects & inputs in order to trigger an autosave. However, all .dropdowns will trigger a "change" even as part of the init (see link above). I do not see how that is helpful, so I propose skipping this on initial load.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2112,10 +2112,10 @@ $.fn.dropdown = function(parameters) {
                 module.add.optionValue(value);
               }
               module.debug('Updating input value', value, currentValue);
-              $input
-                .val(value)
-                .trigger('change')
-              ;
+              $input.val(value);
+              if(!module.is.initialLoad()) {
+                $input.trigger('change');
+              }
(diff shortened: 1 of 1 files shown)


jlukic:
I'm not seeing the `onChange` event being called on initial load in any of my test cases
http://jsfiddle.net/y62xgj7L/

Can you provide an example dropdown where this occurs?

lauri-elevant:
I forked your JSFiddle with an update:
http://jsfiddle.net/3atkecbh/1/

I am actually referring to the "native" change event. My use case is basically:

	$(document.body).on('change', 'input,textarea,select', function() {
		// we need to trigger autosave..
	});
So if I set this functionality up first and then go ahead an instrument a dropdown, I get a change event. I think there was no change, semantically speaking :)

See the files changed. The code actually a triggers an update event.

jlukic:
I've implemented a similar fix for `2.1`

lauri-elevant:
Thanks! Much appreciated.

Issue #1878, opened 2015-02-26
Which is the proper way to work with npm modules and browserify?


trigoesrodrigo:
Hi. I'm trying to use some of the SUI modules inside an application in which I'm already using Browserify and npm modules.

I found some problems while trying to get the modules to work.
So far I've tried four modules: semantic-ui-checkbox, semantic-ui-dropdown, semantic-ui-tab and semantic-ui-modal.

With semantic-ui-checkbox everything worked properly with the following code:

```
var $ = require('jquery');
var Checkbox = require('semantic-ui-checkbox');
$.fn.checkbox = Checkbox;

....

//In my Marionette.View onShow method
this.$el.find('.ui.checkbox').checkbox();
```

However, with the other three modules I've encountered problems:

* semantic-ui-dropdown: just requiring it produced an error with browserify that goes away if I remove the underscore in this [line](https://github.com/Semantic-Org/UI-Dropdown/blob/8047fbfa25888e57a7cba87714b5fe1f6169e3dc/index.js#L1741), probably a bug. After fixing that, another error occurs because internally it is using the transition module. I've tried including the module before (without the need of manually doing `$.fn.transition = require('semantic-ui-transition')` because this time the module does it automatically -I think either all modules should do that or none, otherwise it could be misleading-). After trying that, the error is different and  `Uncaught ReferenceError: _module is not defined` pops up in this [line](https://github.com/Semantic-Org/UI-Transition/blob/aaf0fa2517d0ee777c94224fae7ea58f12425556/index.js#L508) after I click the menu.
* semantic-ui-tab: similar problems as semantic-ui-dropdown. In this case `Uncaught TypeError: Cannot read property 'exports' of undefined` pops up in [line](https://github.com/Semantic-Org/UI-Tab/blob/52024be09ab913de5d01c727bf93f3810b30d7fe/index.js#L28) when I invoke `.tab() on the DOM elements.
* semantic-ui-modal: apparently this module depends on dimmer and dimmer on transition, so I've tried requiring them before (in the same fashion as before `$.fn.dimmer = require('semantic-ui-dimmer')`. After doing that, `Uncaught ReferenceError: _module is not defined` pops up in [line](https://github.com/Semantic-Org/UI-Transition/blob/aaf0fa2517d0ee777c94224fae7ea58f12425556/index.js#L508) when I invoke the modal. It looks as the same error as with the dropdown module.

Am I doing something wrong? Are there any examples of using that modules in a browserify environment?

[...]

PR #2816, opened 2015-08-09
Update RegExps in packaging for npm.

davide:
Fixes npm packaging for components api, site (#1878) and transition.

This patch is re-implementation of fixes done in pull request 1156 (https://github.com/Semantic-Org/Semantic-UI/pull/1156) following some concerns raised in #1878.

### DIFF
--- a/tasks/admin/components/create.js
+++ b/tasks/admin/components/create.js
@@ -93,8 +93,8 @@ module.exports = function(callback) {
             unrelatedNotes    : new RegExp('^((?!(^.*(' + component + ').*$|###.*)).)*$', 'gmi'),
             whitespace        : /\n\s*\n\s*\n/gm,
             // npm
-            export            : /\$\.fn\.\w+\s*=\s*function\(parameters\)\s*{/g,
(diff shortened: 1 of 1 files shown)


jlukic:
What modules have you checked the modification against? 

I know specifically some modules [use a different syntax](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/behaviors/api.js#L16) for extending jquery.

davide:
I went through all repos (generated by gulp release) that had an index.js file in there, comparing the files before and after the regex replacements.
The new RegExp/replacement retains whatever assignment is already in place (which was the case in the site and api modules):
```
$.api = $.fn.api = function(parameters) {
```
turns into:
```
var _module = module;
$.api = module.exports = function(parameters) {
```

jlukic:
I'll release a patch tomorrow

jlukic:
@davide Do you think it makes sense to leave `UI-Transition` out of `dependencies` since this is not actually required by modules `index.js`, or should it be left in?

jlukic:
Tagging on npm as we speak, let me know if that resolves issues.

PR #2832, opened 2015-08-13
add projecttalk to readme

manukall:
I've noticed you have to tag a lot of issues with "Read the Contributing Guide / Readme" or "Should Use StackOverflow / Gitter".

This PR adds a badge and links to the Semantic-UI message board on ProjectTalk and hopefully redirects these questions there.

I only recently started ProjectTalk. If you consider this PR Spam, please excuse and let me know. But I think it's genuinely useful.

I also noticed you already use google groups, but I think ProjectTalk has a number of benefits over that. For one, having the badge in your Readme, it's just more visible. Also, people won't need a google account. And most importantly, it's just more beautiful, as it's built with Semantic-UI ;-)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ### Usage Questions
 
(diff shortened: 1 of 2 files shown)


jlukic:
I'm cool with this.

PR #2836, opened 2015-08-13
Fix callback return from build task

artemkaint:


jlukic:
Why didn't I do this months ago... so obvious

Thanks

jlukic:
Ah, check readme, PR should be merged into `next` branch. Will manually merge commit.

PR #2839, opened 2015-08-13
Use hexadecimal literals istead of octal literals in install tasks

artemkaint:
In JSLint the warning  "Don't use octal: '{a}'. Use '\u...' instead" is given
In JSHint and ESLint the warning has always been "Octal literals are not allowed in strict mode"


jlukic:
Added in 065709b

PR #2843, opened 2015-08-14
Tabs does not work properly for duplicate names

habibutsu:
Example of bug: http://jsfiddle.net/kby77myL/1/


jlukic:
Looks good, thanks!

PR #2869, opened 2015-08-18
Single step border radius fix

elliottisonfire:
There's an issue with the single step element where the border radius style is adhering to a "last child" case. You can see the bug here: http://jsfiddle.net/o7wn9xjd/. I am able to replicate in IE10, IE11, Chrome, Safari and Firefox (I did not test the mobile browsers).

The solution addresses the concept of "the latter rule wins". In this case, the rule for `:last-child` holds true during any `:only-child` case. This was overriding the `:only-child` style.

### DIFF
--- a/src/definitions/elements/step.less
+++ b/src/definitions/elements/step.less
@@ -102,11 +102,6 @@
   border-radius: @stepsBorderRadius 0em 0em @stepsBorderRadius;
 }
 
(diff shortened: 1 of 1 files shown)


jlukic:
Thanks for the clear test case and fixed. Added, but removed nested rules.

PR #2886, opened 2015-08-20
Add support for transition none to the popup module.

msarris:
I'm working on support for IE9 for a site, and for this I need support for transition none in the popup module. This was already available in the dropdown, which is implemented about the same way.

I was required to put the call to module.bind.close(); in a timeout callback to bind it in the next event loop iteration. This is because otherwise the popup would close immediately due to the click event still being bubbled.

I have another suggestion: it would probably be a good idea to default to no transition when transition is not supported. This way you will still support the maximum browsers, and won't compromise the code with hacks or anything.

### DIFF
--- a/src/definitions/modules/popup.js
+++ b/src/definitions/modules/popup.js
@@ -375,7 +375,15 @@ $.fn.popup = function(parameters) {
         animate: {
           show: function(callback) {
             callback = $.isFunction(callback) ? callback : function(){};
-            if(settings.transition && $.fn.transition !== undefined && $module.transition('is supported')) {
+            if(settings.transition == 'none') {
+              module.set.visible();
+              $popup.transition('show');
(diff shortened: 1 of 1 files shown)


jlukic:
You can also do this with existing code using `duration: 0` and any transition.

msarris:
That would give the same effect for browsers that support the transition module, but browsers that don't will already fail on the $module.transition('is supported') call.

I see you added the pull request to the 2.2 milestone, thanks a lot. Keep up the good work, this is a great project. :)

jlukic:
I'll merge in with next minor. Thanks for the code.

lulalala:
Is this still going to be merged? Thanks!

jmezach:
Any updates on this? This would help us out a lot.

PR #2892, opened 2015-08-22
[Form] Remove Defunct Fluid Form

razzius:
Noticed this was missing.

* Not sure if it needs an `!important` but at least in the docs
it wasn't getting overwritten by anything.
* Also not sure if the compiled css files are
supposed to be included (they aren't in this PR)


jlukic:
Form should be 100% width by default. 

Could you provide a test case where this is necessary?

razzius:
Turns out it I was using `.grid` incorrectly. See http://jsfiddle.net/f7575jm4/4/

I ended up realizing that the grid in this case was unnecessary - the form layout by itself was sufficient.

It might be worth removing `fluid` from [form variations](http://semantic-ui.com/collections/form.html#fluid)  if forms are fluid by default though.

jlukic:
Yeah that shouldn't be there. Removing.

jlukic:
(clicked wrong button) 

Updated docs to remove fluid from form

PR #2902, opened 2015-08-25
[Icons] Upgrade to Font Awesome 4.4

ImSanchez:
I'm new to GitHub. I would like to help so I went ahead and added the new icon classes from Font Awesome 4.4. I'm going to better organize them later if this is added.

### DIFF
--- a/dist/semantic.css
+++ b/dist/semantic.css
@@ -7367,7 +7367,7 @@ i.icons .inverted.corner.icon {
 }
 
 /*
- * Font Awesome 4.3.0 by @davegandy - http://fontawesome.io - @fontawesome
+ * Font Awesome 4.4.0 by @davegandy - http://fontawesome.io - @fontawesome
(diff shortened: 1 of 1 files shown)


jlukic:
Hi, thanks for starting work on this. 

The files in dist/ are automatically generated, you'll need to modify 'themes/default/elements/icon.overrides'.

Generally each icon has a primary name. Aliases are at the end of the file. Make sure to keep the order in overrides which is by UTF content. This helps with updates

ImSanchez:
Ah okay, thanks.

kappaknight:
One recommendation: within the Currency section of this page - http://semantic-ui.com/elements/icon.html you can fill the last empty spot with the Bitcoin icon.

kappaknight:
What about a credit card version of Bitcoin? (Similar to VISA, MC, Discover, Amex, etc.)

jlukic:
Check other PR above.

Issue #2779, opened 2015-08-04
[Search] - Ajax abort not correct "there was an issue querying the server."


ayurmedia:
**Steps to Reproduce**
1. use searchfield with .ui.search and remote api
2. type some searchwords
3. quickly change the searchwords to trigger an ajax abort
4. it starts a new api search
5. it shows server-error because of the bug 

**Expected**
The Search should do this: 
1. Enter word, start search via ajax (ok)
2. Enter another word, abort ajax (ok)
3. Start another request (ok)
4. show results of the new request (broken)

**Result**
The Search does not do this: 
1. The search shows server error, but should show result
2. ajax abort does not work on instance but globally on module.request

Bug is somewhere module.request.rejectWith, which will break ajax requests after this
$api.search.query.event.fail...
if one returns in fail() then search works like expected. 

**Testcase**
the bug can be reproduced on the semantic-ui page: 
http://semantic-ui.com/modules/search.html
![screen shot 2015-08-04 at 18 15 45](https://cloud.githubusercontent.com/assets/1746990/9065905/e9b633e0-3ad4-11e5-9e86-74fdc6d7e361.png)

when you quickly type different words in the searchbox you can produce "abort". (unfortunately semantic-ui-com often responds with 502 Bad Gateway, but it should be abort. 

the GET is Aborted, and another GET is successful. 
the API Debug now should trigger the display of the results, but because of global vars or a race-condition the second Ajax request also shows API => "API request failed" which is the cause of the Bug. 
the error-message of the first GET gets copied over to the second GET (ajax) request and then shows the server-error, which is wrong, because the second request was successful. 

if the callback of $api.search.query.event.fail: function()... returns early then the result-list will be shown. 

to turn on debug log in js-console on semantic-ui website enter following in the javascript-console. 

$('.ui.search').search({'debug':1,'verbose':1,'cache':0});

dontfeedthecode:
The problem is easier to reproduce with a remote data source, it doesn't even really need to be typed that quickly. I've got a search component hooked up to a remote API and this error shows up pretty much any time you make a spelling mistake.

jlukic:
Abort is on purpose, this prevents a previous request which takes longer from loading its results after a later request. 

Without abort:
1) "S" - starts loading S results
2) "Se" - self clearing timeout prevents search until user finishes typing..
3) "Sea" - User stops typing, 'Sea' begins querying,
4) "Sea" results load quickly and change the results to results for "Sea"
[...]

PR #2914, opened 2015-08-26
bugfix for ajax abort

ayurmedia:
simply skip the reject for abort will produce the expected behavior. See Issue #2779 .

### DIFF
--- a/dist/semantic.js
+++ b/dist/semantic.js
@@ -18367,7 +18367,7 @@ $.api = $.fn.api = function(parameters) {
               }
               setTimeout(function() {
                 if( module.is.abortedRequest(xhr) ) {
-                  module.request.rejectWith(context, [xhr, 'aborted', httpMessage]);
+                  //module.request.rejectWith(context, [xhr, 'aborted', httpMessage]);
                 }
                 else {
                   module.request.rejectWith(context, [xhr, 'error', status, httpMessage]);


jlukic:
We can't remove the onFailure conditions without preventing `onAbort` from ever firing. There needs to be a less 'brute-force' solve than this.

ayurmedia:
sure you can find a better solution. at least you now acknowledge that the bug is real. the onAbort is not a problem and is the correct way that the ajax api is reacting. the only problem is that the rejectWith triggers an chain of errors in the promises and then the suggest breaks. the correct way is to ignore anything after the abort, as the promiss will not be fulfilled as there is no callback for success, because it was aborted (canceled). the promis chain now calls all fail() methods because the ajax request was stopped. the wrong thing in the logik is that abort is an error, but its not, its only a user-interaction and must not prevent the suggest to show in the future. but the real bug is that the errorcode is bound to the api.js and not the ajax-request. the bug is causing the following ajax request to break, which is the real problem here. the quickfix is only a workaround to allow the correct behavior, because the errorhandling is broken.

jlafleche:
@jlukic any ideas on a better solution for this problem? I'm also seeing this bug, so as far as I can tell the search widget is broken without @ayurmedia's hack.

I'm happy to put together a PR with a fix if you can share your thoughts on what would be an acceptable solution.

ayurmedia:
@jlukic  " We can't remove the onFailure conditions without preventing onAbort from ever firing. There needs to be a less 'brute-force' solve than this." the problem is its not an "error", the onAbort is a User-Action , same as a klick on a link. This wrong understanding is causing a chain of problems. your understanding and architecture is wrong. my bugfix only shows where the root of the problem is.

jlukic:
I don't really understand what's going on here. All I can see is you've short-circuited the onAbort event from firing in this PR.

I'm just hoping for two things
1) A specific test case that consistently reproduces the issue, so I can see what I'm suppose to be triaging.
2) Not breaking API for people who attach callbacks to `onAbort`

ayurmedia:
@jlukic 

for 1) the error is 100% reproducible here: http://php-ayurmedia.rhcloud.com/semantic/search.2.0.8.html see the lower part of #2779 with a dummy echo-server to make the problem more clear. 

its not meant to be just comment out the onAbort or similar like i did in the pull-request, but to actually fix the semantic-ui API. because the onAbort triggers some errorhandling, but it should only ignore the situation as the next request is already coming in a few milliseconds later. also the onabort should be connected to the ajax-thread and not the whole suggest search. eg. one auto-complete field, many ajax requests, but only one global handling of onAbort. (its attached in semantic-ui to the wrong object, otherwise the onAbort would only cancel one of many ajax requests and be ok). 

for 2) the usage of onAbort is wrong, its not an backend-error but an front-end event, like "onClick, OnHover, onSubmit etc. in 99.9% one can simply ignore this event. simply like old browsers can ignore onHover and nothing breaks. 

the only thing you "could" send to the user, "onAbort => "msg: the user decided to do something else, wait for the next event and -do not interfere-". 

the error only appears if you have a slow server and manage to send onAbort and a new request in the timeframe the server needs to response. if you have a search on the API url and the database needs > 1 second response-time this bug is very obvious. if your server reacts in 0.001sec then its hard to reproduce. 
this bug is only visible in a cornercase which is the default-case on a slow server. one reason this bug is open since many many weeks now. 

i use my "bugfix" in a productive webserver as its the correct behavior. this is not just a patch and "hope nothing breaks". 

but its only a workaround to a deeper problem in the semantic-ui API and handling of onAbort in general. there is a chain of event-handling in the API, and my fix is the easiest (but not cleanest) way to force the correct behavior. 

the onAbort should be attached to a single xhr object which there can be many in parallel and not to the autosuggest-object (one searchfield). 

the root of this is in the onAbort of img-tag. where the user can abort the loading of an image by pressing the (X)-Stop button while loading. simply ignore it. its not an error. 

simply "onError != onAbort".

woodlandhunter:
I'm having this issue too (it really happens all the time, even in the docs, etc.) but I do'nt think this is the correct fix. I made a comment here:

https://github.com/Semantic-Org/Semantic-UI/issues/2779#issuecomment-188909108

I also linked to a self contained node/express script to reproduce the issue easily:
https://gist.github.com/anonymous/4b108f8cb6a1ba4f85c9

jlukic:
@woodlandhunter Thanks for the gist link, this will be very useful. I'll fix with next patch release.

jlukic:
See response in #2779. 

I will also take some time today to make sure this fix adequately addresses API uses in dropdown.

PR #2922, opened 2015-08-27
[Input] Support labels attached on both sides

### LINKED ISSUE #2871: [Input] Labeled Should Support Attached on Both Sides
One thing I'm trying to do which seems to not be possible with SUI, is to add both a left and right attached label to an input. Currently only left, or right is possible unless I'm missing something. See here:

http://jsfiddle.net/nz8smghx/1/

I've tried playing around with various classes but it all ends up much the same. One label is fine, the other label is facing the wrong way. I completely understand why it's happening, just would be useful to have the option of adding both

maturano:
fix #2871

Example: https://github.com/Semantic-Org/Semantic-UI-Docs/pull/112

### DIFF
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -339,13 +339,21 @@
 }
 
 /* Regular Label on Right */
-.ui[class*="right labeled"].input  > input {
+.ui[class*="right labeled"].input > input,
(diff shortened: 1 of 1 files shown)


jlukic:
This appears to remove word order deps

maturano:
Yes. I think in code is more clear :smile:

PR #2929, opened 2015-08-31
[Nag] Fix issue with localstorage and add support for session storage

Neaox:
Fixed an issue in Nag module > storage.remove which was checking settings.storageMethod with an incorrect value and was also checking window for a store property which does not exist.

Added support for sessionstorage to Nag module to enable nags to be shown once per session.


jlukic:
Thanks, I appreciate the fixes. I still need to write docs for nag.

PR #2944, opened 2015-09-02
add form data as onSuccess callback param

guodong:
easy to access form data when using ajax submit in onSuccess


jlukic:
Would have to be added as a new parameter to avoid breaking code.

jlukic:
Please check readme these PR should be in `next` branch

PR #2956, opened 2015-09-03
Fixed bad template url

AgentShark:
Equals after the question mark caused parameters not to be recognised by youtube.


jlukic:
Just in time.

PR #2973, opened 2015-09-07
Remove right icon margin

icharlie:
right icon margin

http://jsfiddle.net/icharlie/2kduqf0n/1/

rightIconMargin causes there is a little space on the right side.

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -591,9 +591,6 @@
   vertical-align: @iconVerticalAlign;
   color: @iconColor;
 }
-.ui.button > .right.icon:not(.button) {
-  margin: @rightIconMargin;
-}
 
 /*******************************
            Variations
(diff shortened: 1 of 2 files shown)


jlukic:
That margin is because icons with right in the name, `right icon`, `right chevron` normally appear on the other side of the text. However it shouldnt be affected `labeled icon` button.

jlukic:
Added as a different solve.

icharlie:
Just in case, the margin I deleted is in the `right arrow icon` not the `labeled icon` button

jlukic:
It does, the margin is on the `.labeled.icon > .button`

icharlie:
The code I removed is following
```
-.ui.button > .right.icon:not(.button) {
-  margin: @rightIconMargin;
-}
```

jlukic:
You can see this fiddle from next. This solves both the case of a right arrow button with and without labeled icon. The fix you provided would break the second case.
http://jsfiddle.net/9fcra8c5/

icharlie:
Should the second case looks like?

```
<button class="ui labeled icon button">
  <i class="right arrow icon"></i>
  Next
</button>
```

icharlie:
Maybe, this should be in the docs as an example!?

PR #2979, opened 2015-09-08
[form] Get value of checkbox if its have

### LINKED ISSUE #2954: [Checkbox] - Checkbox do NOT passing value when form is submited
**Steps to Reproduce**
1. Create form with checkbox:
<div class="ui checkbox">
		<input type="checkbox" name="test" value="MyTestValue" checked="checked">
		<label>Test</label>
	</div>
2. Add javascript to submit form $('.ui.form').form({...});
3. Submit

**Expected**
The Checkbox should pass - MyTestValue

**Result**
The Checkbox passing true or false - depend if checkbox is checked or not

**Testcase**
http://jsfiddle.net/rduvhn8u/98/

maturano:
fix #2954 

Related commit 688c645 ... keeps returning false :)

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -455,7 +455,7 @@ $.fn.form = function(parameters) {
                   }
                   if(isCheckbox) {
                     if(isChecked) {
-                      values[name].push(true);
+                      values[name].push(value || true);
                     }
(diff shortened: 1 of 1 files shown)


jlukic:
Thanks @maturano for picking up the pieces there.

Issue #2981, opened 2015-09-08
[Dropdown] Use <select> readonly and disabled properties


mixerp:
The official SemanticUI documentation site suggests replacing plain old HTML select to SemanticUI dropdowns. On MixERP project, we have faced multiple issues that restrict us from using Semantic UI dropdowns as a replacement of HTML select.

## Dropdown Does not Preserve State

Let me give you an example here:

[https://jsfiddle.net/79aos51b/](https://jsfiddle.net/79aos51b/)

After converting to SemanticUI dropdown, the disabled select element should also become disabled dropdown.

However, we can live with that. But the next issue is really troublesome.

## Dropdown Should Behave Like Select

This issue has been already been discussed before and has not been addressed. To remind you again, the empty option is lost once you make a selection. This is wrong. On the other hand, adding a button and triggering "clear" event is a not practical solution at all. Why would a user have to click a button to clear the selection if a wrong selection was made?

veke:
I think too, that Dropdown should behave just like select tag. Disabled property should disable converted dropdown too. In mean time, adding disabled class to select tag do the trick (if this is possible in your case)

I noticed that option without value attribute makes it re-selectable like in normal select.  
https://jsfiddle.net/af32e63c/

maturano:
I make a PR about the disabled state. I made this cases: https://jsbin.com/rapopus/edit?html,output

About the empty option, I think we need more feedback. I personally like the current behavior.

mixerp:
@veke, I did not know that, thank you.

@maturano, the current behavior breaks existing javascript logic if one wants to upgrade selects to dropdowns for a simple reason that it does not work like HTML select.

Your PR looks great for the disabled case, but it needs to apply to **readonly** state as well. :+1:

jlukic:
Currently select can be disabled with `ui disabled dropdown` on the `select` before init. 

However currently, disabled and readonly are not being read from the `<select>`'s properties. I could see this as an enhancement.

With regard to reselecting placeholder, simply initialize with `placeholder: false`, see accompanying conversation in other threads about the thought here.

mixerp:
@jlukic, ```placeholder:false``` I missed the documentation, my bad. Thank you.

maturano:
@mixerp 
> Your PR looks great for the disabled case, but it needs to apply to readonly state as well. :+1:

[...]

PR #2983, opened 2015-09-08
[dropdown] Preserve initial disabled state

maturano:
ref: #2981 

Just add disabled state if the origin is disabled.

```html
<select class="ui dropdown" disabled>
</select>
```
to
```html
<div class="ui dropdown selection disabled">
</div>
```

test: https://jsbin.com/rapopus/edit?html,output


jlukic:
Clear

PR #2997, opened 2015-09-10
fixed broken override for site.variables

SeoFood:
It is not possible to override the site.variables anymore because the line will always called after the override because @theme == @element.

### DIFF
--- a/src/theme.less
+++ b/src/theme.less
@@ -18,9 +18,6 @@
 /* Packaged site.variables */
 @import "@{themesFolder}/@{site}/globals/site.variables";
 
-/* Component's site.variables */
-@import "@{themesFolder}/@{theme}/globals/site.variables";
-
 /* Site theme site.variables */
 @import (optional) "@{siteFolder}/globals/site.variables";
 


jlukic:
You are misinterpreting this. `@theme = @@element;`, notice the two `@@`

This evaluates per-element differently, but assuming `button`, 
```
@element = 'button' //defined in definition
@button = 'github';  // defined in theme.config

// theme.less
@theme = @@element;
@theme = @button;
@theme = 'github';
```

The order for site theme includes an extra level because it uses your specified `@site` theme as well.

* Default theme site.variables
* Selected theme's site variables
* Current component's site variables
* Your site theme's site variables

SeoFood:
According you have the following theme config:

```
@site       : 'mysite';
@icon       : 'default';
```

if you override the @fontPath it will load the vars from default because in the icon less file, because:

`@import (optional) "@{themesFolder}/@{theme}/globals/site.variables";` == `@import (optional) "themes/default/globals/site.variables";` 

So if you want to override the fontPath you need to add it to the _site/globals/site.variables or use `@icon       : 'mysite';`

PR #3042, opened 2015-09-16
[Search] Close search results after setting the input value.

ahtinurme:
I'm trying to clear the input value from onResultsClose with $(search).search('set value', ''), but that gets overwritten since the callback is executed BEFORE the value is set internally and no other callback is available to do it after setting the value internally.

This pull request is to fix the issue.


ahtinurme:
One way would be  to do it with setTimeout, but that's not the most elegant way.

PR #3046, opened 2015-09-16
Allow <table> styling with any tags via classes .table, .tbody, .td, etc

jayphelps:
Applying the table style currently depends on you actually using a real `<table>` and its descendants.

```html
<table class="ui table">
  <thead>
    <tr>
      <th>Header</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>Cell</td>
    </tr>
  </tbody>
</table>
```
Proposed opt-in/typecast class names:

```html
<div class="ui table">
  <div class="thead">
    <div class="tr">
      <div class="th">Header</div>
    </div>
  </div>
  <div class="tbody">
    <div class="tr">
      <div class="td">Cell</div>
    </div>
  </div>
</div>
```

Pros: Works with any number of infinite scroll table components that do not use a real `<table>`.
Cons: Increases overall filesize about 7 KB minified.

Flame on. :fire: :fire:

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -50,15 +50,19 @@
 
 /* Table Content */
 .ui.table th,
-.ui.table td {
+.ui.table td,
+.ui.table .th,
+.ui.table .td {
   transition: @transition;
 }
 
 /* Headers */
-.ui.table thead {
+.ui.table thead,
+.ui.table .thead {
   box-shadow: @headerBoxShadow;
 }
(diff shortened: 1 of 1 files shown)


jayphelps:
The usefulness of this ended up being more questionable than I thought. 💩

jlukic:
I was withholding judgment. I know it does feel a bit weird that this is one of the two (icon) components that rely on tags for meaning. But this one is pretty inescapable.

Issue #3045, opened 2015-09-16
Form validation issues


miniskulljob:
Hi.

The new form validation shorthand has some issues. Check out [this JSFiddle](http://jsfiddle.net/miniskulljob/Lcfg0rv0/). In it, I have used the two methods: the conventional one, and the shorthand. Conventional works fine, but the shorthand always returns a success, even if the field is empty.
Another issue is that both methods call two times the result function, `onSuccess` or `onFailure`.

jlukic:
Yeah... I'm not sure why this obvious case isnt working.. grr...

maturano:
I made a PR that I believe solve this issue[*]. You can see the result here: https://jsbin.com/xahide/3/edit?html,console,output

[*]: Although I don't know what I'm doing :sweat_smile:

miniskulljob:
Yeah, seems like it works, but it isn't using the official repos from semantic.org, maybe there's something different...
And, in your jsbin, the result function is called two times too.

jlukic:
I'll try to take a look at this soon.

maturano:
@miniskulljob 
> but it isn't using the official repos from semantic.org

That's right. Is using the version of my fork, just for test the patch.

> And, in your jsbin, the result function is called two times too

It seems that something has changed at this time. Also your JSFiddle change; both cases doesn't works, now generate the error `TypeError: range.indexOf is not a function` (confirm please)

This PR solve this new error... I keep watching

federicotarantino:
Hi,
related to this issue, i suggest implementing range logic also on "decimal" rule.
In my project I create a "range" rule that work with integer and decimal.

jlukic:
I believe this was fixed but mistagged.

PR #3053, opened 2015-09-17
[form] Fix integer validation

maturano:
fix issue #3045


jlukic:
Thanks this is clear.

PR #3073, opened 2015-09-22
Global utilities

phparkle:
I've used Semantic UI on several projects and I found that I always ended up patching these styles to site.overrides. I thought it would be better if there could be a separate component for global utilities. Some of the styles are already implemented in certain components, for example `.ui.right.floated.image`. However, having the global utilities means that you can put `class="ui right floated"` on any element, not just images, and it will right float it. Similarly with the responsive visibility classes, right now they are only implemented in ui.grid but with the utilities you can easily apply responsive visibility to any element. I find this extremely useful when prototyping. I think there could be other useful utilities as well, this is just a start. What do you think?


zuck:
I agree they are very useful. In Bootstrap I used them a lot.

CodeAnimal:
I totally agree, having utility classes seems like a no-brainer to me. It could also have the added benefit of condensing the code base as repetitive code (like having float classes implemented for specific components) could be mostly collated into one class, excepted for exception cases that may need extra styling.

Having utility classes would still enable semantic styling while adding extra usefulness to the framework as a whole, and it would make transitioning from Bootstrap (and other frameworks) that little bit easier and more attractive. :+1:

kamranayub:
I was going to open an issue for this--I've always been frustrated at these "generic" utilities not being available for all components. I would also love to see typographic utilities, such as aliasing the header tags to classes to easily apply font size changes to elements quickly. All you need is to change the globals from `h1 { }` to `h1, .h1 { }` etc.

sstok:
@kamranayub That would defeat the entire purpose of having "semantic" HTML 🙀

kamranayub:
@sstok Well you have to be pragmatic too. If I want to keep a hierachical layout for my headings, but want to display the font differently depending on the context, I have to create helper utilities anyway. Not to mention if you need to display sizes differently on mobile vs. tablet vs. desktop, etc sometimes it doesn't make sense to use heading tags when it's easier to manipulate classes via Javascript (like React/Angular). A dynamic UI might need different displays and the utilities can be helpful.

Anyway, that's beside the point of this issue--really it'd be great to document any shared utilities on the Semantic site or pull up any utilities that can be shared across components when there's no reason they need to be specific to a component.

PR #3078, opened 2015-09-23
new brand color for ok.ru 

newsiberian:
Hi, [ok.ru](https://en.wikipedia.org/wiki/Odnoklassniki) is very popular social network in Russia.
Why I did this PR? It's for Oauth plugins which use SUI, for example in Meteor...

BTW, icon for ok.ru will be provided with FA 4.4


jlukic:
I don't believe I'll be adding any additional button styles.

PR #3088, opened 2015-09-24
Adding support for data-text when using dropdown with API requests

johnschult:
See: http://semantic-ui.com/modules/dropdown.html#specifying-different-text--hidden-input-values

There was no way to include `data-text` when using `dropdown` and populating it with the API. This will conditionally add it if it is contained in the response results.

Additionally, the `values` variable was not being used.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3386,10 +3386,7 @@ $.fn.dropdown.settings.templates = {
 
   // generates just menu from select
   menu: function(response, fields) {
-    var
-      values = response.values || {},
-      html   = ''
-    ;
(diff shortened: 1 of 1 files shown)


jlukic:
Thanks I'll have to review

jlukic:
Hi, since this was not merged with `next` (as mentioned in the readme) there appear to be some merge conflicts. 

I've manually merged these in, unfortunately i accidentally committed this with other changes. 
https://github.com/Semantic-Org/Semantic-UI/commit/2b2187c415d34ce586663fab1fe9455cfa171003#diff-dbb1b936fdad1f63b3a01f86d9a54fd1L3666

PR #3096, opened 2015-09-25
[Search] Replaced usage of `response` variable by `result` variable to set URL on result item

akinsella:
Hi,

It seems the wrong variable is used to build and affect associated URL to results in search module. Without that change, it is impossible to associate a specific URL per result. Because other values are found on result variable it seems logical to use the choosen field on `result`variable and not on `response` variable.


jlukic:
It looks pretty obvious just from scanning the code that this is wrong.

fabienb4:
See #3206 for complete fix.

jlukic:
Added in #3206

PR #3136, opened 2015-10-05
Break validation on first error

listepo:
@jlukic see https://jsfiddle.net/b21sgvkk/1/
rule ```emailExists``` running if rule ```email``` not valid and call many ajax requests on blur


listepo:
@jlukic please review this PR

listepo:
Ping

jlukic:
It needs to run through all errors in order to create validation, you'll just need to check for conditions to abort XHR in your emailExists rule.

listepo:
@jlukic thanks

Issue #3106, opened 2015-09-28
onSearchQuery isn't called if search results are cached


mnquintana:
Steps to reproduce:

1. Initialize search module with `throttle` of `100`
1. Search for a query (let's say it was `javascript`) - `onSearchQuery` is called
1. Clear the search input
1. Search for the same query again
1. `onSearchQuery` is not called

I would expect `onSearchQuery` to be called anytime a *query* is made, not anytime a network request is made. If it would make sense, there could be two callbacks - `onSearchQuery` for when the query changes, and something like `onRequest` for when a new remote request is made.

Looks like this is expected behavior judging from this conditional: https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/search.js#L427

jlukic:
It should be called in both conditions

PR #3138, opened 2015-10-05
Call onSearchQuery whether or not results are cached

mnquintana:
Fixes https://github.com/Semantic-Org/Semantic-UI/issues/3106


jlukic:
Thanks @mnquintana

PR #3161, opened 2015-10-11
Inconsistency with `widescreen only` class

mdehoog:
The grid `row`/`column` classes use the `widescreen only` selector, however in the parent `grid` class only `widescreen` is present (missing `only`). Also, in the mobile only hide, the `row`/`column` `widescreen only` is missing, with `large screen only` repeated twice (probably simply a forgotten find/replace after a copy/paste).

Additionally switches the tablet only class selector on the mobile only hide from `.tablet:not(.mobile).only` to `[class*="tablet only"]:not(.mobile)`, for consistency with the surrounding selectors.

### DIFF
--- a/src/definitions/collections/grid.less
+++ b/src/definitions/collections/grid.less
@@ -1790,7 +1790,7 @@
 
 /* Mobile Only Hide */
 @media only screen and (max-width: @largestMobileScreen) {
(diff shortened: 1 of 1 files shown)


jlukic:
I've confirmed the mobile rules are repeated for widescreen, and the other fixes mentioned.

PR #3167, opened 2015-10-12
Support for dimmer with zero opacity

mdehoog:
If the dimmer has a opacity setting of 0, the CSS background-color is set to `rgba(0, 0, 0, undefined)`, due to the way the opacity variable is set. This PR accounts for an opacity setting of 0.


jlukic:
Thanks

PR #3168, opened 2015-10-12
Ability to cancel a modal hide from the onHide callback

mdehoog:
Currently the modal allows the user to cancel a hide from the `onDeny` and `onApprove` callbacks, but not from the `onHide`. This could be useful if you want to stop the user from hiding a modal but don't want to set `closable` to false.

For example, if you have a popup that submits a request to the server, and you want to display a loading state after the user has submitted, you could temporarily disable closing the popup by returning false from the `onHide` callback while the form submission is in progress.

Happy to also create a PR for the docs repository documenting this change.

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -365,7 +365,10 @@ $.fn.modal = function(parameters) {
             : function(){}
           ;
           module.debug('Hiding modal');
-          settings.onHide.call(element);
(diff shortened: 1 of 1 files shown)


jlukic:
I don't see any downsides to this.

mdehoog:
Thanks @jlukic. PR to update the docs for this change has been submitted.

You've done a fantastic job with this library.

PR #3179, opened 2015-10-14
Improve dropdown search performance

parisholley:
If you have a large number of elements in the dropdown, using search causes the browser to stall out. This change removes some of the jQuery object allocation that seems to cause the issue.


jlukic:
Check PR notes in readme for proper file locations. I've gone ahead and re-implemented. 

I haven't noticed massive performance gains profiling this with < 100 choices, but generally it seems like it should be faster.

PR #3180, opened 2015-10-14
Fixed inconsistent use of media query breakpoints

mdehoog:
In `modal.less` and `step.less`, a `max-width` of `@computerBreakpoint` is used, causing an inconsistency when the browser is exactly 992px width (ie most queries switch on the 991-992 change, but these switch on the 992-993 change).

- `@largest*` should only be used by `max-width` queries
- `@*Breakpoint` should only be used by `min-width` queries

This PR makes the use of these variables more consistent.

### DIFF
--- a/src/definitions/elements/divider.less
+++ b/src/definitions/elements/divider.less
@@ -132,7 +132,7 @@
 }
 
 /* Inside grid */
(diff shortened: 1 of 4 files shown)


jlukic:
I can confirm these should all be @largest(X)Size as you've specified. Merging in.

PR #3217, opened 2015-10-20
Make checkbox labels cursors always look consistent

davialexandre:
While doing usability tests in an app that is using Semantic-Ui, some users got confused by the lack of consistency of the checkbox label's cursors. When the javascript module is in use, the pointer cursor is always used. When it isn't, the cursor is text or default, depending on the checkbox state.

With this PR, I made it consistent in every scenario (with or without the javascript module):
- When the checkbox is active, the pointer cursor will be used
- When the checkbox is disabled or read-only, the default cursor will be used

Here is a fiddle that shows how the labels are currently inconsistent: http://jsfiddle.net/davialexandre/w24cehL6/

### DIFF
--- a/src/definitions/modules/checkbox.less
+++ b/src/definitions/modules/checkbox.less
@@ -63,7 +63,7 @@
 
 .ui.checkbox .box,
 .ui.checkbox label {
-  cursor: auto;
+  cursor: pointer;
   position: relative;
   display: block;
   padding-left: @labelDistance;
@@ -231,7 +231,8 @@
 ---------------*/
 
 .ui.read-only.checkbox,
-.ui.read-only.checkbox label {
+.ui.read-only.checkbox label,
+.ui.checkbox input[readonly] + label {
   cursor: default;
 }
 
@@ -261,7 +262,6 @@
 
 /* Selectable Label */
 .ui.checkbox input.hidden + label {
-  cursor: pointer;
   user-select: none;
 }
 


kvetoslavnovak:
+ 1

jlukic:
The correct solve here is probably to only use cursor: pointer on unitialized checkbox if there is a `[for]` property on label

jarrettv:
Agree this needs to be fixed. When it looked broken I just wrapped the radio button in a label and removed semantic from this part of the form.

henriquebremenkanp:
:+1:

sebilasse:
:+1: 
But the main problem I am seeing here are the 
`.ui.form .inline.field>label`
and
`.ui.form .field>label`
rules.

They prevent the checkbox labels to inherit the user's `label` variables because with the above markup
```html
<div class="field">
  <div class="ui checkbox">
    <input type="checkbox" name="checkbox" id="checkbox" value="1">
    <label for="checkbox">Checkbox</label>
  </div>
</div>
```
the label is *not* an immediate child of .field but of .checkbox...

Please see this [fiddle](http://jsfiddle.net/xe1fzvca/13/) –

do we "need" the `"for"` markup at all? An alternative :
```html
<label class="inline field ui checkbox">
  <input type="checkbox" value="1">
  <label>checkbox</label>
</label>
```
would do the same in terms of usability and [accessibility](https://24ways.org/2015/the-accessibility-mindset/)

But to support the "label wrapper approach" it needs a block default:
```css
.ui.form .field {
  display: block;
}
```

and btw: As you can see from the fiddle I am also working on crossbrowser modules for input types `range` and `color` - I'd like to contribute them soon...

larsbo:
I think this could be easily fixed with the css I mentioned here: https://github.com/Semantic-Org/Semantic-UI/issues/3355#issue-117526972?

Issue #3225, opened 2015-10-22
.attached.segment within a .aligned grid doesn't have proper width


jholl:
here's a [codepen](http://codepen.io/jholl/pen/rOJLZO)
using default theme

the styles add `width: 100%` to all `.segment:not(.compact)` within variations of `.aligned` grids/columns.  This overrides the necessary `.attached.segment` declaration of `width: calc(100% + 2px)` to account for the `margin: 0em -1px` on `.attached.segment`

jlukic:
Fixed with #3226

PR #3226, opened 2015-10-22
Don't override .attached.segment width

jholl:
Addresses issue #3225


jlukic:
Please submit PR against `next` branch

PR #3277, opened 2015-11-02
fix stackable grid children with dividers inheriting parent behavior

andreivolt:


mdehoog:
This breaks the following use case:
```
<div class="ui stackable grid">
  <div class="row">
    <div class="column">
      column1
    </div>
    <div class="ui vertical divider">
      or
    </div>
    <div class="column">
      column2
    </div>
  </div>
</div>
```
Currently it is not a requirement to mark a `row` as `stackable` if the `grid` is `stackable`. Do we need to add another selector with `row` without `stackable` to all these cases, or am I wrong in putting a non-`stackable` row inside a `stackable` `grid`?

mdehoog:
After some testing, adding `stackable` to `row` in a `stackable grid` also fixes #3187. Maybe a non-`stackable` `row` inside a `stackable grid` is not such a good idea after all. Will add `stackable` to my `stackable grid` rows from now on.

Issue #3260, opened 2015-10-29
Unknown flag not handled


ayazmaroof:
When using the Flag element from Semantic UI we found that if an unknown or non-existent country code was used then Andorra's flag is displayed by default. There should be a better way of handling this e.g. an icon/flag with a question mark.

PR #3301, opened 2015-11-08
Fail safe flag without valid country code class

space-alien:
A potential fix for #3260. Render blank instead of showing the flag of Andorra when the country code class name is missing or invalid.


jlukic:
Thanks this seems clear.

space-alien:
I have created a new PR on the next branch:

https://github.com/Semantic-Org/Semantic-UI/pull/3333

PR #3333, opened 2015-11-15
Fix for flags with invalid country class

space-alien:
Fix for #3260. When a flag has a missing or invalid country class name, render blank instead of showing the flag of Andorra.


jlukic:
Thanks @space-alien , all set

Issue #3364, opened 2015-11-20
If form serialization is used in api, it extends formData forever


tvildo:
Hi
I am using this code to initialize form and attach api:

``` javascript
$filterForm.form({
    fields: {
      ...
    }
}).api({
    defaultData: false,
    action: 'get history',
    serializeForm: true,
    beforeSend: function (settings) {
        // open console to inspect object
        console.log(settings.data);
        return settings;
    }
});
```

Api extends formData in each query call instead of assigning serrialized form values;
**Problem :**
   when checkboxes  are serialized on the form if they are unchecked keys are removed from js object  as you know. so if data is extended 4ever unchecked checkboxes will stay  checked 

blame is this line on api.js

``` javascript
 // Add form content
          if(settings.serializeForm) {
            settings.data = module.add.formData(settings.data);
          }
```

here it  formData body : 

``` javascript
formData: function(data) {
            var
              canSerialize = ($.fn.serializeObject !== undefined),
              formData     = (canSerialize)
                ? $form.serializeObject()
                : $form.serialize(),
              hasOtherData
            ;

https://github.com/Semantic-Org/Semantic-UI/issues/new#/* this line is same as : data = settings.data || settings.data */
            data         = data || settings.data; 
            hasOtherData = $.isPlainObject(data);

            if(hasOtherData) {
              if(canSerialize) {
                module.debug('Extending existing data with form data', data, formData);
                data = $.extend(true, {}, data, formData);
              }
              else {
                module.error(error.missingSerialize);
                module.debug('Cant extend data. Replacing data with form data', data, formData);
                data = formData;
              }
            }
            else {
              module.debug('Adding form data', formData);
              data = formData;
            }
            return data;
          }
        }
```

settings.data = formData(settings.data) 
which always  extends data again:

``` javascript
if(canSerialize) {
                module.debug('Extending existing data with form data', data, formData);
                data = $.extend(true, {}, data, formData);
              }
```

wenweih:
Structured form data requires including [macek's serialize object.](https://github.com/macek/jquery-serialize-object)

PR #3365, opened 2015-11-20
fixes checkbox serialization in form 

tvildo:
If form serialization is used in api, it extends formData forever, we should serialize form and assign result to data  #3364


jlukic:
I dont understand the issue.

wenweih:
@jlukic Any progress about this pr ?

wenweih:
@tvildo  a message in Semantic-UI Doc  
> Structured form data requires including macek's serialize object.

PR #3396, opened 2015-11-27
[Modal] Support modification of 'closable' setting while modal is open

mdehoog:
If the `closable` property of modal is modified while the modal is visible using `$('.ui.modal').modal('setting', 'closable', value);`, it does not take effect for mouse events until it is hidden and shown again. This is because the `closable` setting affects whether a click event is bound, rather than checking the setting in the click event handler itself.

Note that changing `closable` for a visible modal works for the escape key, because the key is checked in the event rather than a conditional binding.

This PR makes the behavior for click consistent with the escape behavior, by binding the click event whenever the modal is visible, and checking whether to close within the click handler itself. This means that the `closable` setting can be changed while the modal is visible.


piya23300:
+1

mdehoog:
Fixed merge conflicts, would really love this one to be merged. Thanks @jlukic

lulalala:
Bump this, and can we mark this as a bug so it gets more attention?

DornaHansen:
+1

Jake-Lippert:
+1

jlukic:
Merging with small modification

PR #3432, opened 2015-12-03
[Dropdown] Allow ignoring of case when getting an item

Neaox:
Currently get.item compares labels and values case sensitively. This can cause duplicates when `allowAdditions` is `true`. To prevent these duplicates get.item should be capable of comparing labels and values while ignoring case.

However the current behaviour should be preserved by default.

To achieve this a new setting is added called `ignoresCase`, which when `true` will disregard the casing of labels and string values when performing the comparison.


jlukic:
Thanks, I'll take a look

micahbule:
Any updates on this one?

Banandrew:
@micahbule The current milestone set is 2.2.8, which is the next framework version. I understand that it was postponed several times, but hopefully it will make it into 2.2.8.

micahbule:
@Banandrew right. Totally understandable. I made a work-around for our project for now. Thanks for the quick response.

alytle:
This would be really nice to get in 12 if possible

jlukic:
I think this looks good as is.  I'm going to update setting to be `ignoreCase` and get in this patch

jlukic:
From testing this fix doesnt appear to work with current SUI.

jlukic:
I've implemented this in 0fa5821, the solution was slightly more involved because `$.inArray` is not case insensitive.

PR #3463, opened 2015-12-12
Fix for search selection dropdown

tpflueger:
When using search selection dropdown, the `selected` class is removed from the actual selected item and is placed on the first item of the list. Removed filter code from removing and selecting the first item in the list.


jlukic:
The removed code prevents selection from being left on a filtered item. I'm not sure why you would want to remove it.

TylerPflueger:
I created issue 3444 with fiddle. If you select an item within the dropdown, that selected item gets removed and the filtered selected item is always kept on the first item within the dropdown.

Issue #1462, opened 2014-12-17
[Build] Port Karma Test Runner


Wexcode:
Simply tried to run `karma start` after following install instructions, and got the error in the title. After searching around for a bit, this article seemed the most promising: 

http://stackoverflow.com/questions/22367059/no-provider-for-frameworkjasmine-resolving-frameworkjasmine

I can't seem to find any documentation here - could somebody shed some light on the best way to get this set up?

PR #3490, opened 2015-12-19
Adds karma from gruntfile #1462

tpflueger:
Brings karma over into gulp task to run through tests.


jlukic:
The karma tests havent been updated for 2.0, so this would probably not yield any benefits.

TylerPflueger:
I can understand there is no benefit currently, but it kinda bothered me seeing outdated tests and nothing that could show things are working as expected. If there is another issue out there to do so, I would like to look into it.

PR #3492, opened 2015-12-20
Add a style option

sadorlovsky:
It allows some tools like `npm-css` and `postcss` to import styles from the node_modules directory. It's awesome to write just `@import "semantic-ui"` to load css.


jlukic:
Thanks, I'll take a look into how those packages function.

Issue #3514, opened 2015-12-28
The checking for semantic.json always fails when npm dependent package uses Semantic UI.


Jeff-Tian:
My project requires another project, say `online-store`, which uses `Semantic UI`.  I need to automate the  build process of my project and added a `gulp task`. 

The problem is when I run that `gulp task`, it always claims that the `semantic.json` can't be found, while it actually indeed exists in the `online-store` project that under `node_modules` folder.

![image](https://cloud.githubusercontent.com/assets/3367820/12014428/33ac26bc-ad67-11e5-86e1-08b6860ce401.png)

After investigation it caused by the `when.hasConfig()` method in `install.js` didn't pass any context folder to the `requireDotFile()` method, which in consequence fallbacks to use `__dirname` to start searching `semantic.json`.

So the `install.js` should pass a context folder, say `process.cwd()` to the `hasConfig()` method:

![image](https://cloud.githubusercontent.com/assets/3367820/12014445/9f960172-ad67-11e5-8938-8d30c8bd7478.png)

jackjwilliams:
@Jeff-Tian I think you may have solved a few issues.

#2056 
#3244 (possibly)
#2821 
#2841 (Could be related - not sure, it is a pathing issue)

I just started connecting some dots when I saw other [seemingly related] issues. Hope it helps.

PR #3515, opened 2015-12-28
#3514: the fix for issue: 'The checking for semantic.json always fail…

Jeff-Tian:
…s when npm dependent package uses Semantic UI'.


Jeff-Tian:
The checking for semantic.json always fails when npm dependent package uses Semantic UI. #3514

jlukic:
Thanks I'll check this out

PR #3542, opened 2016-01-07
vertically attached dropdown

larsbo:
add styling for top attached dropdown and bottom attached dropdown
(similar to attached buttons)

example fiddle: http://jsfiddle.net/rr3Ltdzh/


jlukic:
Hi, you might also want to verify other attachments, as these get a bit complex.

larsbo:
Yes, I will test the other elements & modules in the next time. For my use case I just needed the segment attached version and thought it might be a good extension for the framework. Thus I created the PR to find out if this is useful at all.

Banandrew:
From what I’ve tested—I copied the example above and tried appending dropdowns to the elements—it works as expected. The only problem is when an attached dropdown is placed between other attached elements like tables or segments, there’s a double border issue, which, I believe, can be fixed with something like that:
```CSS
.ui.attached.table + .ui.attached.dropdown:not(.bottom),
.ui.attached.segment + .ui.attached.dropdown:not(.bottom),
.ui.attached.header + .ui.attached.dropdown:not(.bottom) {
  margin-top: -1px;
}
```
There’re an issue with attached menus being offset by 1px (#4248), maybe it’s better to deal with it first. Also, the fork was updated from the original, some merge bubbles appeared, and since it doesn’t exist anymore, this PR can’t be updated to remove them.

larsbo:
thank you for validation, @Banandrew :)
I deleted the branch because I had no time for checking all the other attached elements and the pr didn't get much attention at all. Moreover the pr was cluttered with non related modifications as you already mentioned. 

If desired, you can use my changes for an own pr 😃

Issue #2994, opened 2015-09-10
[Item] Image not centered in Item View in IE11


pantosha:
In IE11 image isn't centered.
See picture:
![2015-09-10_110218](https://cloud.githubusercontent.com/assets/1799798/9783473/20335106-57ad-11e5-8835-27d4439200bf.png)

PR #3584, opened 2016-01-18
Fix image not centered in Item View on Mobile in IE11 #2994

pantosha:
Now image is centered. But in IE width of Item element more than width of window.

image


pantosha:
Create new pull request in `next` brach https://github.com/Semantic-Org/Semantic-UI/pull/3818

PR #3586, opened 2016-01-18
replace doesn't work on undefined

pablo-coco:
these two lines have to be switched.


Benew:
is no one using rtl ? why is this still an issue ?

PR #3596, opened 2016-01-21
Fixes left arrow key in dropdown search input

Sanjo:
For the case where no item is selected. Because in this case $parentMenu is still something else than $menu.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1184,20 +1184,21 @@ $.fn.dropdown = function(parameters) {
 
                 // left arrow (hide sub-menu)
                 if(pressedKey == keys.leftArrow) {
-
-                  isSubMenuItem = ($parentMenu[0] !== $menu[0]);
-
-                  if(isSubMenuItem) {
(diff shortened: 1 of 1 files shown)


jlukic:
Can you explain what might trigger this issue?

jlukic:
@Sanjo Delaying this without a specific test case

Sanjo:
Reproduce:

1. Go to the "Search Selection" example on http://semantic-ui.com/modules/dropdown.html.
2. Type "foo" into the search input
3. Press the left arrow key to change the cursor position in the text. This currently doesn't work and this fix fixes that.

This fix just adds the condition:

```
if(hasSelectedItem) {
  ...
}
```

which is necessary because `isSubMenuItem` has the value true even when no item is selected, which interprets the left arrow key press wrongly as menu navigation action, instead of the default cursor position change action.

jlukic:
Woah that is bad.. thanks for the catch

jlukic:
Thanks for spotting this, this is a pretty heinous bug. 

I've manually added the fix instead of merging due to conflicts when merging into the next branch. (See readme for details, but generally all PR should be merged into `next` branch)

PR #3598, opened 2016-01-21
Embed: only add question mark when parameters are set.

happylinks:


jlukic:
This is cleaner, merging into `next`.

Issue #1816, opened 2015-02-18
[Build] Add Quiet Option


mendenhallmagic:
First, I just wanted to say thanks for this promising new release. Your dedication to this project is amazing and inspiring. I'm really excited to simplify and speed up our build using the new semantic.less file...

I just have one obstacle to upgrading:

My team has a jenkins build process, so running running npm install with a fresh (empty) node_modules directory is a use-case we have to support. Since the new release runs gulp install without a default semantic.json, user interaction seems to be required to advance the process. The way I got around this before was to npm install, copy a default semantic.json from our project root into the node_modules/semantic-ui, copy the semantic-ui/src/_site folder to the semantic-ui/src/site folder, copy site.variables from our project into semantic-ui/src/site/globals, then run gulp build.

Is there currently a way to run npm install without requiring user interaction?

jlukic:
It probably wouldnt be too difficult to add since build scripts have [config data](https://docs.npmjs.com/misc/scripts#packagejson-vars) available. 

Would just require modifying `package.json` to include some extra flag.
https://docs.npmjs.com/files/package.json#config

Will brainstorm on this

kapouer:
seconded ! i also note that installing semantic-ui now pulls shitloads of gulp modules - this is not so friendly, as not everybody is keen using gulp.

mendenhallmagic:
Based on my initial testing, it looks like npm config variables do not cascade, so this may not be an option :/

Is there a reason why semantic-ui runs gulp install automatically? Users of the gulp build system would probably be alright with taking the extra step to run gulp install themselves...do you agree?

jlukic:
@kapouer I will be creating separate repos/distributions for LESS/JS only and CSS/JS only, these will shadow the main repo.

@mendenhallmagic The issue is they'd have to run `gulp install` from inside `node_modules/semantic-ui `which is unusual.

jlukic:
@mendenhallmagic 

It's worth noting that if theres a `semantic.json` in your project root, NPM will skip install questions.

kapouer:
@jlukic thank you for the semantic-ui-css and al. repos ! great !

jlukic:
I think having a `semantic.json` in your project root should solve all issues with needing `quiet` version of install, @mendenhallmagic can you confirm you have everything working?

mendenhallmagic:
[...]

PR #3616, opened 2016-01-27
Add an automatic reinstallation process based on semantic.json

algorithme:
This feature is added to allow to automatize a semantic-ui
re-installation with the same configuration (such as in your CI build)

To switch it on, simply add ("auto-install": true) in semantic.json

Of course, configuration name could be changed :)

### DIFF
--- a/tasks/config/project/install.js
+++ b/tasks/config/project/install.js
@@ -416,7 +416,7 @@ module.exports = {
           {
             name: 'Yes',
             value: true
-          },
+          }
         ]
       },
       {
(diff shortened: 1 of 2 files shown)


algorithme:
Could work as a workaround for some cases quoted in https://github.com/Semantic-Org/Semantic-UI/issues/1816

jlukic:
This seems to be an important issue, i'll try to take a look soon.

dsnoeck:
@jlukic any news here ? Is there anything we can do to get it merged ?

jlukic:
Looking at this today

jlukic:
Thanks for your patience, fixed issue where autoInstall was returning `semantic.json` instead of setup questions, modified style to match.

PR #3632, opened 2016-01-31
Next

byrebv:
Why not Progress with data-minim settings built-in, working close with limitValues ? ... 

`<div class="ui small indicating progress" data-value="210" data-minim="200" data-total="300" id="pret-rez">`


jlukic:
This is incorrect PR

Issue #1736, opened 2015-01-30
[Search] Api search using custom ajax call


Trouche:
Firstly, I have really been enjoying Semantic UI.  It saves me a LOT of time during the build.  Our application is being built in Ember and Semantic is working great within that environment with little or no modifications.

However, today I finally came up against an Ember-specific problem with Semantic.  I would very much like to use the Search module to fetch small amounts of data from the server.  I understand that the API is the recommended way of retrieving that data.  However, while Ember does use the $.ajax method of retrieving data, it doesn't use the typical jQuery implementation but rather an Ember.$.ajax one.  Additionally, all data calls are typically wrapped in an Ember promise to ensure that it is correctly added to the Ember run queue.  The Semantic API abstracts the actual ajax call.  Short of hacking the API (which I really don't want to do), how can I manually specify how I fetch my data?  As we may be searching 9000+ records, pre-caching is out of the question.  

Clearly, this is beyond the scope of typical usage, but any assistance you can provide would be greatly appreciated.  Providing the search with a third way of getting content (local, API, custom), even if it was more complex for the developer, would make the Search module much more flexible for application developers.

Thanks,
John Pendergrast
RocketRez (www.rocket-rez.com)

jlukic:
So my understanding is that calls to `$.ajax` in ember are done through `Ember.$.ajax`? Can you provide a link for documentation on this for me within Ember?

I think i can provide abstractions for this inside `$.fn.api` that could detect for different libraries and call the appropriate internal method.

jlukic:
In previous/current versions can retrieve an API promise using 

```javascript
var promise = $('.ui.search').api('get request');
```

I've also just addded in `next` a new callback `onRequest` that is called when an `XHR` request is made, it receives the `promise` and `xhr` object as 1st and 2nd params.
```javascript
$('.ui.search').search({
 apiSettings: {
  onRequest: function(promise, xhr) { 
    // do something fun with this
  }
  }
});
```

Trouche:
I am using a very customized data model with Ember (written by guys close to the Ember team).  As such, the only way that I could use the Semantic API would be to be able to manage my own data calls.   This, of course, defeats some of the primary purpose of the API.  In the interim I have leveraged other pieces of Semantic and created my own lookup (search) implementation.  It isn't nearly as pretty or full featured as yours, but it works for now.  Thank you for looking into this issue.  I look forward to the solution you develop.  :)

aaronbhansen:
[...]

PR #3653, opened 2016-02-04
[API] onResponse was not obeying return dataType

bockscar:
I'm new to Semantic and trying to get brain around a few concepts. Thank you for this awesome work.

I am exploring using Semantic with a legacy backend. The legacy backend returns JSON that requires processing. One key in the returned JSON is 'html' from which, say a tab's HTML contents, would be pulled.

Though my question applies to all API calls, running with tabs is a useful case study. A tab is hardwired to seek and get returned from server plain HTML. My objective would be to leverage as much Semantic tab logic as possible but get it to work with this legacy JSON backend.

This discussion regarding a related topic discusses the use of 'mockResponseAsync' for a related purpose:
https://github.com/Semantic-Org/Semantic-UI/issues/1736

It seems onResponse is the simpler solution. I was glad to find it. But I did not understand why the response text was being transformed into a hash before being sent to the 'onResponse' callback. The first thing I need to do on the other end is reassemble the *original* response text so I can process it in proprietary fashion. Maybe there's a reason for this transformation?

My suggestion retains the default and offers a flag to skip this wasteful transformation so it needn't happen every server hit if it is not necessary.

Maybe I misunderstood 'mockResponseAsync' and that's where I should be working? If not, maybe the default behavior for 'onResponse' transforming the text might be confusing to others as well.

Thanks for the awesome work. I look forward to getting deeper. Even without this change, it's cool that hooks are in place so I can wire Semantic into a legacy app.

### DIFF
--- a/src/definitions/behaviors/api.js
+++ b/src/definitions/behaviors/api.js
@@ -469,7 +469,7 @@ $.api = $.fn.api = function(parameters) {
                 elapsedTime        = (new Date().getTime() - requestStartTime),
                 timeLeft           = (settings.loadingDuration - elapsedTime),
                 translatedResponse = ( $.isFunction(settings.onResponse) )
-                  ? settings.onResponse.call(context, $.extend(true, {}, response))
+                  ? settings.onResponse.call(context, (settings.onResponseSendPlainText ? response : $.extend(true, {}, response) ) )
                   : false
               ;
               timeLeft = (timeLeft > 0)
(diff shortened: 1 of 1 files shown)


jlukic:
If you could create a text case that reproduces the specific issue with `onResponseAsync` it'd be easier to talk about.

bockscar:
Sorry. Wanted to include some info on the use case if it happened to apply to anyone else. The callbacks actually do exactly what we need (aside from this minor friction).

Heart of the question was understanding why `response` is passed **after** being converted to an object. Other places `response` is passed as a string. That's why this usage stood out and confused me initially...because I was expecting a string:
`settings.onResponse.call(context, $.extend(true, {}, response))`

jlukic:
It looks like this might be a bug when `dataType` is not specified to the default JSON. I'll have to take a look.

I don't think we need a separate setting, it should just mimic the expected response type.

jlukic:
I've taken a different angle, but this should fix it as long as you are specifying a different `dataType` than JSON or JSONP

Issue #3192, opened 2015-10-15
[Modal] { inverted: true } works only on the first run.


alexandretperez:
$('.ui.modal').modal({  inverted: true }).modal('show');

Shows the modal with white background. After you close it and then open it again the background is now black.

Run Code twice at "Dimmer Variations" section to see the bug:
http://semantic-ui.com/modules/modal.html#/examples

jlukic:
This is a bug related to doc initialization (the modals are re-used in all the examples), but I don't believe a bug with SUI?

alexandretperez:
I did some changes and now it works.

Was need instantiate the modal only once.

Here is the test case:
http://plnkr.co/edit/RnUqli4JnYIuxs9v88YC

larsbo:
I got the same unexpected behavior. Here is a common use case with an modal which lost it's inverted state after fired once: http://jsfiddle.net/8x15wobw/
The same result with ```dimmerSettings```:
```js
dimmerSettings: {
	variation: 'inverted'
}
```

brunotourinho:
same problem here, even in the Examples pages the bug occurs:

> http://semantic-ui.com/modules/modal.html#dimmer-variations

on the first call to inverted modal, the dimmer is clear, if you make a second call, the dimmer goes dark.

avalanche1:
@larsbo , you should not instantiate your modal on every click.
http://jsfiddle.net/8x15wobw/1/

PR #3665, opened 2016-02-08
fix modal/dimmer inverted bug #3192

larsbo:
During the second dimmer run the color was extracted from the DOM. At
least Firefox automatically converts RGBA to RGB if possible which
caused the bug.

With this fix the dimmer module can now detect both RGB and RGBA colors.

### DIFF
--- a/src/definitions/modules/dimmer.js
+++ b/src/definitions/modules/dimmer.js
@@ -383,10 +383,11 @@ $.fn.dimmer = function(parameters) {
             var
               color      = $dimmer.css('background-color'),
               colorArray = color.split(','),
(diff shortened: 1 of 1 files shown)


jlukic:
Thanks, this is clear

lubber-de:
pure rgb values are still wrongly fetched, because the last array element still has its closing parenthesis , so an existing 
```
rgb(1,2,3)
```
 and an opacity setting of `0.4` results in
```
rgba(1,2,3),.4)
```
We fixed this in Fomantic-UI by https://github.com/fomantic/Fomantic-UI/pull/1561

Issue #3622, opened 2016-01-29
[Dropdown] item data value is cleared when dropdown message is shown


mdehoog:
When creating a searchable dropdown from existing HTML, the dropdown item's `value` property in the element data is cleared when searching and no results are available. This is because, when the "No results" message is shown, the `menuObserver` mutation observer is fired which calls `module.refresh()`, which in turn calls `module.refreshData()` which removes the `value` and `text` data values from the item elements.

Here is a jsbin that reproduces the issue: http://jsbin.com/runetopasi/edit?html,js,output

Here is the code that removes the `value` and `text` data from the items: https://github.com/Semantic-Org/Semantic-UI/blob/04a3c48a0dd3bd9b440b59036004409239a6db16/src/definitions/modules/dropdown.js#L385-L387. Could you please explain the reason behind this code? Otherwise, simply removing these lines will fix this issue.

jlukic:
See #3879 for response

PR #3680, opened 2016-02-11
[Dropdown] - Placeholder text is lost after dropdown refresh

kevspadilla:
Remove the line that removes the placeholder text fixes the bug.


jlukic:
Can you provide a test case. It should be recaptured during init
```javascript
 // dropdown line 93
 // remove placeholder
 module.refreshData();
 // set placeholder to correct value
 module.save.defaults();
```

mdehoog:
Also see #3622. What is the reason for this reset, as well as the reset of the text/value data?

PR #3763, opened 2016-02-26
Fixes #3761 issue with misspelled callbacks

### LINKED ISSUE #3761: [Checkbox] onEnabled/onDisabled setting misspelled or regressed
I managed to track down the cause of the bug. It is a regression that was introduced by fixing #3148. I don't think it's possible to fix this and avoid the backwards incompatibility unless both forms of the methods are defined as settings and called inside enable.

[Reproduced bug jsFiddle](http://jsfiddle.net/1jqfpgtm/1/)

petruavram:
This is a possible fix for #3761 but I don't know about the possible backwards compatibility issues. I do not think is prudent to have both versions of the callback in the enable/disable methods.

### DIFF
--- a/src/definitions/modules/checkbox.js
+++ b/src/definitions/modules/checkbox.js
@@ -268,7 +268,7 @@ $.fn.checkbox = function(parameters) {
           }
           module.debug('Enabling checkbox');
           module.set.enabled();
-          settings.onEnabled.call(input);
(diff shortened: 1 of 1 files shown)


petruavram:
The issue was fixed by ed98c55e8471ce27a9c76305f5fb2448d05e11e3 by @jlukic

PR #3795, opened 2016-03-07
Fixes #3794 - Dropdown issues a burst of requests for first autocomplete word

### LINKED ISSUE #3794: Dropdown issues a burst of requests for first autocomplete word
When typing in the first word in `search` dropdown, the API issues a request for each letter because the test whether the API is initialised always fails. This slows down the performance and adds many unnecessary overlapping requests.

bez4pieci:
Fixes #3794. 

The test whether the API has been set up always fails if throttle parameter is greater than `0`. The `get request` behaviour returns false because due to throttling no request has been created. 

My proposed test for the presence of an initialised API solves this.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -680,9 +680,12 @@ $.fn.dropdown = function(parameters) {
               }
             }
           ;
-          if( !$module.api('get request') ) {
+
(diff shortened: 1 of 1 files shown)


jlukic:
This has already been fixed in `next`, check issues linked in #3794

PR #3797, opened 2016-03-08
Fix .disabled.reveal

sapjax:
As the [document](http://semantic-ui.com/elements/reveal.html#disabled) says:
>  A disabled reveal will not animate when hovered

Maybe the css selector in reveal.less is wrong.

### DIFF
--- a/src/definitions/elements/reveal.less
+++ b/src/definitions/elements/reveal.less
@@ -219,17 +219,7 @@
               States
 *******************************/
 
-.ui.disabled.reveal:hover > .hidden.hidden.content {
-  position: static !important;
-  display: block !important;
-  opacity: 1 !important;
-  top: 0 !important;
-  left: 0 !important;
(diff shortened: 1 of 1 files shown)


jlukic:
This would make the `visible content` `display: none;` I'm, we want the opposite right?

sapjax:
@jlukic 
I think `visible content` should be `display:none`, if `disabled` just mean disabled the animate.
Any way the origin code is incorrect，set `display:none` on the `hidden content` make no sense.

jlukic:
Visible content is the content that is initially visible. When a reveal animation is disabled it shouldnt show the `hidden content`.

sapjax:
@jlukic I got it!, so the `disabled` is not only means **disabled the animation**, It also means **reset the state to begining**.
But the `hidden content` is initially hidden, why we make it invisible when hovered?  i'm confused.

Issue #2405, opened 2015-06-15
[Search] Remove English Text from Message Template


tomitrescak:
Hi

I am trying to localise the search results, so that it will display "No results" in appropriate language. But, passing following doesn't do anything:

``` javascript
$(".search").search({ source: values, errors: { error: { noResults: 'Custom text ...'}} });
```

jlukic:
Your settings object is just a little bit off. Here's the correct init.

```javascript
  $(".ui.search").search({
    source: values,
    error: {
      noResults: 'Custom text ...'
    }
  });
```

furins:
I'm re-opening this since I have the same issue and I can provide a test case:

source HTML:

    <div class="ui search elements">
        <input type="text" class="prompt" name="element" value="">
        <div class="results elements" ></div>
    </div>


Javascript code:

    $('.ui.search').search({
      source: values,
      error: {
        noResults: 'Custom text ...'
      }
    });

generated HTML when searching for non-existing value:

     <div class="ui search elements focus">
         <input type="text" class="prompt" name="element" value="" autocomplete="off">
         <div class="results elements transition visible">
             <div class="message empty">
                 <div class="header">No Results</div>
                 <div class="description">Custom text ...</div>
            </div>
        </div>
     </div>

I'd like to change the "No Results" message, not only the `.description` text. How can I do it?

thanks!

jlukic:
Can you provide a JSfiddle?

furins:
sure, here it is: 

https://jsfiddle.net/xnfkLnwe/

jlukic:
I should probably move the string out of the template.
https://github.com/Semantic-Org/Semantic-UI/blob/next/src/definitions/modules/search.js#L1158

You could technicaly fix this now by overwriting $.fn.search.templates.message, but it'd probably be easier just to add a new string

furins:
Thank you so much! 
Do you mean $.fn.search.**settings**.templates.message, am I right?

to future readers: here is a proof of concept of jlukic suggestion, until 2.0 will be released https://jsfiddle.net/xnfkLnwe/1/

jlukic:
Yes I meant 

```javascript
$.fn.search.settings.templates.message=  function(type, message) {
  html = '<div>some html</div>';
  return html;
};
```

radcliff:
btw, this line should have the arguments in the other order:

` message: function(type, message) {} `

 should be

` message: function(message, type) {}`

if i'm not mistaken

[...]

PR #3803, opened 2016-03-09
Issue #2405. Fix static "No Results" message.

0x3333:
Fix issue #2405.

The "No Results" message is static in the template. This fix created a property in the error array in the settings.

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -883,8 +883,8 @@ $.fn.search = function(parameters) {
         displayMessage: function(text, type) {
           type = type || 'standard';
           module.debug('Displaying message', text, type);
-          module.addResults( settings.templates.message(text, type) );
-          return settings.templates.message(text, type);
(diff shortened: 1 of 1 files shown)


jlukic:
Makes sense to include, but I think we should use a `text` settings object like other modules.

0x3333:
I'll take a look and make the change.

0x3333:
When you say text settings, you mean this: https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/behaviors/form.js#L1062 ?

jlukic:
yes, instead of `noResultsHeader` as a setting,  something like
```
text: {
  noResults: {
    header: 'No Results',
    message: 'Your search returned no results'
  }
}
```

Then instead of passing just the header to the creation function, pass the whole `settings.text` object.

Also keep in mind PR should be merged into `next` branch. Feel free to reopen as separate PR>

bsdo64:
Any progress here?

0x3333:
This can be closed, as PR #4454 has been created.

Issue #3595, opened 2016-01-21
Item Image not responsively centered on iPhone


srissel:
Dear Semantic UI support,

on iOS, images within items are not centered in responsive view.
For reproduction, please have a look at Semantic UI item documentation and open the page on iPhone in landscape orientation.
http://semantic-ui.com/views/item.html#items

The image stays aligned left, where in other browsers and devices, it is centered.

![itemimage](https://cloud.githubusercontent.com/assets/3600737/12479928/cb71d50c-c03e-11e5-85cf-7737d13f0c8c.png)

Is this a known issue?

Kind regards,
Sascha.

srissel:
Dear Semantic UI team,
is there any news on this?

Kind regards,
Sascha.

agilebluefox:
It seems to be working for me after adding 'margin: auto' on the image tag. I can't attest that it works everywhere, but maybe that could be a good starting point to solve the problem.

PR #3818, opened 2016-03-12
Fix image not centered in Item View on Mobile in IE11 #2994

pantosha:
Now image is centered. But in IE width of Item element more than width of the window.

image


agilebluefox:
This patch will also fix issue #3595 in which a similar problem exists on Safari (9.1.1) and iOS (9.3.2).

PR #3828, opened 2016-03-14
Fix typo

enbo:
bordered not bordeblack
(someone probably tried to find and replace 'red' and overlooked this)


enbo:
Sorry, just realized that this was master and not next.

PR #3829, opened 2016-03-14
Fix typo

enbo:
should be 'bordered' not 'bordeblack'

probably an issue with find and replacing the term 'red'


jlukic:
Thanks!

Issue #211, opened 2013-10-09
[UI] Nag


jlukic:
## Nag

Nag is a component for providing important alerts above your website. These alerts can be dismissed and not shown again by a user with the value being stored by cookie, local storage, or api endpoint.

https://github.com/jlukic/Semantic-UI/blob/master/src/modules/nag.js
### Required for 1.0
- Create documentation
- Create examples
- Minor code retooling
### Required for 2.0
- Nag can be automatically hidden after a set number of seconds
- Multiple nags can appear as growl style notifications at the top of the page

dharma-guardian:
The module seems to be broken. The example page http://semantic-ui.com/modules/nag.html throws 'Nag: undefined'.
On OSX 10.10 Chrome v42 / ubuntu 14.04 Chromium v41 + Firefox v37

iamolivinius:
Looks like jquery-cookie plugin is missing. Load it right after jquery and you'll be fine.

jlukic:
It's borked. Probably from when i reworked global includes in the docs. Will reupload tomorrow.

jlukic:
Main docs should not be broken anymore, just reuploaded
http://semantic-ui.com/modules/nag.html

I'll try to write out simple docs for this for 2.0. Probably wont have time to recode it.

iamolivinius:
Just found out the the Nag component has the Popup component as a hard dependency. Without the Popup component the Nag component can not be dismissed correctly. 

I like the idea that components can be loaded separately but I suggest that dependencies should be listed in a subsection of each components documentation.

jlukic:
Nag doesnt have any dependencies. It can use `$.cookie` for storing values, but otherwise none.

iamolivinius:
Without the popup module the `easeOutQuad` function won't be available. Problem: the nag message does not hide.

Without the icon module the `close` icon won't be visible. Problem: the nag message can not be dismissed.

For me, from a users point of view, these are dependencies to get the nag module up and running regarding to the current documentation.

http://jsbin.com/ziweqeqasu/1/edit?html,js,output

jlukic:
Ahh its the damn easing equations. I can fix that.

jpeffer:
Are there any currently scheduled plans to include colors to the nag component? It would make sense if the same styles as the Message collection. Nag really feels very similar to the dismissible Message, with the exception that the decision to dismiss is persisted.

d8vjork:
[...]

PR #3862, opened 2016-03-23
Bake 'easeOutQuad' easing function into nag module.

iamolivinius:
https://github.com/Semantic-Org/Semantic-UI/issues/211#issuecomment-106290794

As mentioned above the nag module is missing the easing function to work properly. The normal jquery library only supports 'swing' and 'linear' easing functions. The added code is a copy from here: 

https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/accordion.js#L589-L593

Please merge and make people happy :tada:

### DIFF
--- a/src/definitions/modules/nag.js
+++ b/src/definitions/modules/nag.js
@@ -483,4 +483,11 @@ $.fn.nag.settings = {
 
 };
 
+// Adds easing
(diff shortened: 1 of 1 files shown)


jlukic:
Thanks for the fix.

PR #3865, opened 2016-03-23
Cards improvement: add cards count per device/screen

DanDrix8391:
I added a way to set a different  amount of cards for each row per device/screen. Just like columns, but now for cards.

for example:
1 card per row for mobile
3 cards per row for tablet
4 cards per row for small screen
5 cards per row for large screen
6 cards per row for widescreen

`<div class="ui one mobile cards three tablet cards four wide computer cards five wide large screen cards six wide widescreen cards">`

### DIFF
--- a/src/definitions/views/card.less
+++ b/src/definitions/views/card.less
@@ -834,6 +834,528 @@ a.ui.card:hover,
   margin-right: @tenCardSpacing;
 }
 
+/*--------------
+   Card Count per device
+---------------*/
+
+/* Mobily Only */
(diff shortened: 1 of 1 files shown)


jlukic:
I'm a bit confused on the mixing of `(x) wide` naming, for computer, and `(x)` for mobile. 

For example `two wide computer`, but `two mobile`.

It would be more consistent with the rest of the library to include `two mobile` and `two computer` (for two cards per row.

brianveltman:
I have the same as @jlukic

PR #3871, opened 2016-03-25
Set Internet Explorer support to 11+

nesk:
Fields are using `pointer-events: none` when disabled with `<div class="disabled field">...</div>`, thus avoiding the children to be clicked. However, this CSS property isn't supported by Internet Explorer 10, [according to CanIUse](http://caniuse.com/#feat=pointer-events).


jlukic:
I think this is fine. IE10 is old news now.

salarmehr:
IMHO it's too soon to drop IE10 support.

PR #3879, opened 2016-03-28
[Dropdown] Removed deletion of text/value metadata

### LINKED ISSUE #3622: [Dropdown] item data value is cleared when dropdown message is shown

mdehoog:
Fixes #3622.

Happy to put this behind a flag instead if you think the current behavior should be the default.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -391,10 +391,6 @@ $.fn.dropdown = function(parameters) {
 
         refreshData: function() {
           module.verbose('Refreshing cached metadata');
-          $item
-            .removeData(metadata.text)
-            .removeData(metadata.value)
-          ;
           $module
             .removeData(metadata.defaultText)
             .removeData(metadata.defaultValue)


jlukic:
It looks like the internal issue was that mutation observers were accidentally triggering `refresh` during internal updates to `addition` or `message`. 

I've resolved this by refactoring the mutation observers to ignore internal changes. 

The reason we have to flush data attributes in `refresh` is that otherwise updating `data-value` (attribute data values) will not trigger the underlying `.data('value')` call to update. 

These values are typically updated when a mutation observer detects the underlying DOM has been rewritten.

jlukic:
Fixed in `2.2`

mdehoog:
@jlukic thanks so much for following this up 👍

Issue #2381, opened 2015-06-08
CSS File Size is Too Large


ianrtracey:
Big fan of Semantic UI so far and are actually looking to implement it for an open source project here at Cisco, but we are quickly finding the initial CSS file too large to comfortably ensure its performance across our user base. Is there anything we can do in regards to segmenting styling with the components actually being used to reduce load times?

ph7vc:
You can use the build tools (gulp) to create a smaller version of the CSS with the components that you want.

By default; it outputs each specific element / collection / view / module CSS file into the `dist` directory.

jlukic:
Semantic UI will continue to grow linearly with each new component added. Currently we're sitting at around fifty components which is much larger than other UI frameworks. 

It's recommended to use the NPM install which will give you a choice for which components to include when installing.
![Getting Started](https://dl.dropboxusercontent.com/u/2657007/install.gif)

Updating this after the fact is as easy as modifying a [`semantic.json`](https://github.com/Semantic-Org/Semantic-UI/blob/master/semantic.json.example) file.

Jeffboq:
The limitation of the current build system is that components not listed in the `semantic.json` file are not included in the `dist` directory.
It could be great to have a way to create a minimal `semantic.css`/`semantic.js` base and also having all components available in the `dist/components` directory.

i8ramin:
It would also be nice to have even more control over what is included in each individual component. For example, the `button` component includes a lot of different button styles that one may never use. Being able to include only the styles needed would be really great. A tool like `UnCSS` (https://www.npmjs.com/package/gulp-uncss) could be used during the build process (specially on a statically generated site), but that is not always possible.

BKD-Digital:
I understand everyone's concern here but this is really a non-issue.

They've already broken the library down to components: https://github.com/Semantic-Org/Semantic-UI/tree/master/dist/components

I advise you take an hour and select which rules you need and eliminate the rest on your own.

jlukic:
:+1:

BillyParadise:
Sorry for commenting on this closed issue.
As an example, I was able to reduce my JS from 395912 to 224060 and my CSS from 554853 to 389632.

A suggestion (I might get to work on this myself some day), the comments at the top of each file should list the dependencies.  The dimmer and transition modules, for example, seem to be required for other modules.

edouard-lopez:
[...]

PR #3928, opened 2016-04-15
Support specifying sections in semantic.json to reduce output size

nix1:
The output CSS file, as described in issue #2381, seems to be way too large for many people.
This merge request adds a way to add specific sections to be taken from the sources in the `semantic.json` file. If the sections are not defined for a component then it will behave the old way, that is the whole component would be included.

Sections are case insensitive and interpreted as regular expressions.
They must be specified in the same order they appear in .less files.

Example of sections configuration for buttons:

```
  "sections": {
    "button": [
      "theme",
      "button",
      "states", "hover", "focus", "down", "active", "disabled",
      "types", "labeled button", "icon",
      "variations", "floated", "compact", "sizes", "icon only", "basic", "labeled icon", "fluid", "primary"
    ]
  }
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "gulp-rename": "^1.2.2",
     "gulp-replace": "^0.5.3",
     "gulp-rtlcss": "^0.1.4",
+    "gulp-tap": "^0.1.3",
     "gulp-uglify": "^1.2.0",
     "gulp-util": "^3.0.6",
     "gulp-watch": "^4.3.1",
--- a/tasks/build/css.js
+++ b/tasks/build/css.js
@@ -8,6 +8,7 @@ var
   // node dependencies
   console      = require('better-console'),
   fs           = require('fs'),
+  path         = require('path'),
 
   // gulp dependencies
   autoprefixer = require('gulp-autoprefixer'),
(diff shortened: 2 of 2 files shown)


jlukic:
This does seem interesting. Can you elaborate on what the regular expression is parsing for.

nix1:
The regular expression looks complex, but it's just parsing for sections starting with a proper block comment opening and having a name. Sections end at the start of the next section.

These are examples of section headers (taken from `button.less`):
```
/*******************************
           Variations
*******************************/
```
```
/*-------------------
       Colors
--------------------*/
```
```
/*--- Red ---*/
```
Note that this is *not* considered to be a section header:
```
/* Basic */
```

jlukic:
I'm glad I used consistent comment headers.

This actually seems like a reasonable/useful feature. I will have to explore the implementation.

nix1:
I'm glad, too :thumbsup: Note that the names are not unique (example: `Columns` occur twice in `grid.less`), which might make it somewhat more difficult to handle. Perhaps you can just rename them.

/cc @larchchen, we worked together on that piece of code

jlukic:
I don't mind changing some comment headers to get this working. Let me know if you see any other duplicates.

lubber-de:
The community fork [Fomantic-UI](https://fomantic-ui.com) solved the problem by providing lots of switch variables for each variant via LESS conditions.
https://github.com/fomantic/Fomantic-UI/blob/develop/src/themes/default/globals/variation.variables

https://github.com/fomantic/Fomantic-UI/pull/781

Issue #3755, opened 2016-02-25
[Validation] Issues with E-mail RegExp


paulborza:
Try this email address `hajdu karcsi@yahoo.com` and it will validate fine according to Semantic UI, but the fact is that email address is invalid.

Morrolan:
@paulborza  Is it validating the full thing or is it validating ```karcsi@yahoo.com``` and dropping the first part?

Can you create an example case in jsfiddle?

paulborza:
@Morrolan Here's the JS fiddle you asked for: [https://jsfiddle.net/x0Lq0s8o/3/](https://jsfiddle.net/x0Lq0s8o/3/)

You can reproduce the bug there per the 3-step instructions I've added.

jlukic:
Thanks for Thanks a lot for filing this issue, and including an easy to understand test case. I've added a milestone, and will try to take a look as soon as I can.

ScopeyNZ:
This is a lot worse than just spaces. It appears to return true if a valid email is at least part of the string. Try `<script>alert('hello');</script>guy@example.co.nz` - returns valid.

It appears the regex is missing the start and end delimiters ('^' & '$'). 

Try `(new RegExp("^[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$", 'i')).test('guy@example.co.nz')`. - I copied the code from the regex used in master.

I haven't really done any further testing on this but that appears to solve the issue.

jlukic:
I'd accept a PR here. Regex's related to emails appear to have been solved in the completely generalized case literally nowhere on the internet.

Ryuno-Ki:
@jlukic Um, there are [RegExes for this](https://gist.github.com/justingit/7012331/), but you don't want to build them by hand. The spec isn't [easy to grasp](https://en.wikipedia.org/wiki/Email_address#Syntax) so I'd suggest doing it the Perl way here.

daneren2005:
[...]

PR #3955, opened 2016-04-22
Email fix for #3755

ScopeyNZ:
In the end I did a bit of research and borrowed the [email regex used in angular.js](https://github.com/angular/angular.js/blob/4487d4a8a477ad440500e3574c141da48d13ab3d/src/ng/directive/input.js#L27).

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1066,7 +1066,7 @@ $.fn.form.settings = {
   regExp: {
     bracket : /\[(.*)\]/i,
     decimal : /^\d*(\.)\d+/,
-    email   : /.+@.+\..+/i,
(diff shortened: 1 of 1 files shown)


kevinresol:
Seems that this regexp is more accurate: 
```js
/^([\w-]+(?:\.[\w-]+)*)@((?:[\w-]+\.)*\w[\w-]{0,66})\.([a-z]{2,6}(?:\.[a-z]{2})?)$/i
```

Examples: http://try.haxe.org/#0B65c

jlukic:
If it works for angular it works for me

 @kevinresol If you can qualify the differences in your regex I would consider.

jlukic:
Added rlsnotes

kevinresol:
@jlukic in the link I posted previously you can press "Build + Run" to try it.

You should be able to notice that `k.@cc.com` is considered as valid email address by angular's regex while it should not be. And my suggested regex correctly recognize that.

PR #3959, opened 2016-04-24
[Search] Add "showNoResults" setting (true/false)

Reikooters:
With the showNoResults: false setting, this suppresses the "No results found" message, so that search can be used simply for suggestions rather than an actual search.

### DIFF
--- a/dist/semantic.js
+++ b/dist/semantic.js
@@ -13074,11 +13074,16 @@ $.fn.search = function(parameters) {
               return false;
             }
           }
-          $results
-            .html(html)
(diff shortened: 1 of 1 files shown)


jlukic:
Hi, this feature has already been added (nearly identical) in next version

https://github.com/Semantic-Org/Semantic-UI/commit/68b007461a257d301b6c606992c3bbec451451f4

Reikooters:
Great thank you for all your hard work. I was using master and needed this. I'll check the next branch in future :)

PR #3966, opened 2016-04-26
let is not supported on Safari

HitomiTenshi:
supported browsers reference: http://caniuse.com/#search=let


jlukic:
Thanks, gotten used to ES6

Issue #4004, opened 2016-05-06
[Form] Label for inline field should inherit inverted form styling


tbracken:
JSFiddle example: http://jsfiddle.net/xykhh7xo/1/

Labels for inline fields do not inherit the lighter (white-ish) color that they should from `.ui.inverted.form`

jlukic:
Closed with #4004

PR #4005, opened 2016-05-06
[ref #4004] added inline fields to inverted form styling

tbracken:
screen shot 2016-05-06 at 4 06 46 pm


jlukic:
Thanks, I was noticing this oversight in my personal projects as well.

PR #4010, opened 2016-05-10
Dropdown menu tpl disabled attribute fix

eymengunay:
When a dropdown menu gets regenerated, menu options loose their "disabled" attribute.

I have added the missing attribute in template string.


jlukic:
I've merged this into `next`, be sure to PR any future requests against that branch (see readme)

PR #4051, opened 2016-05-20
Update material theme font URLs to be '//' instead of 'http://'

AndyR207:
The default material theme uses http:// for the fonts in every override file, which throws warnings on SSL enabled sites. This fixes that.


cweagans:
IMO, switch to HTTPS instead of protocol relative URLs.

jlukic:
Should be `https` like default theme.

AndyR207:
I have forced HTTPS on the URLs now.

cweagans:
:+1:

PR #4075, opened 2016-05-27
Remove extra space

MarcGuay:
Found this while trying to search using copy and paste from the console which must be stripping the extra space


jlukic:
Thanks for another contribution. We only accept PR into next branch (which is a bit ahead of master). See the readme for more details.

PR #4079, opened 2016-05-30
Remove duplicated .grid class selectors

MarcGuay:
Maybe I'm misunderstanding something but it seems like ".ui.grid.grid.grid" is redundant and could be replaced with ".ui.grid"


jlukic:
Since css does not allow you to specify 'specificity' but automatically calculates it based on selectors, in some rare cases, this 'trick' is used to make an `important` rule override another `important` rule by adding additional selector specificity. It is intentional.

Issue #4062, opened 2016-05-26
Undesirable whitespace in .meta


japgolly:
See in http://jsfiddle.net/w95kv4ct/1/
Extra whitespace appears and turns `"ago."` into `"ago    ."`

lixonic:
@jlukic #4095

jlukic:
The whitespace is intended as is, I would recommend wrapping parts of `meta` in child elements.

PR #4095, opened 2016-06-04
No margin-right for children of .meta

lixonic:
Fix for  #4062 

http://jsfiddle.net/lixonic/tfm7qxzz/7/ 

agobug

### DIFF
--- a/src/definitions/views/item.less
+++ b/src/definitions/views/item.less
@@ -213,8 +213,11 @@
   line-height: @metaLineHeight;
   color: @metaColor;
 }
-.ui.items > .item .meta *:not(time) {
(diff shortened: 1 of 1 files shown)


src/definitions/views/item.less
```diff
@@ -213,7 +213,7 @@
   line-height: @metaLineHeight;
   color: @metaColor;
 }
-.ui.items > .item .meta * {
+.ui.items > .item .meta *:not(time) {
```
japgolly: Adding a `>`, so `.ui.items > .item .meta > *` would be better. The `<time>` tag is just one example.

lixonic: Done

PR #4100, opened 2016-06-06
Let onRate know if it's the initial call

krishnaglick:
Setting an initial rating on a rating element calls into onRate. This is an issue within my ember component as this will not be allowed in Ember 3.0 as my onRate function mutates the component's values within the render event. This is probably not the optimal way to handle this, but it allows the user to choose to do something, if at all, due to this event.


jlukic:
This is already a setting in `next`. `fireOnInit`, please wait for `2.2`

PR #4107, opened 2016-06-07
Sets border-radius when vertical buttons have a single child button

Denhai:
Example
```
<div class="ui vertical buttons">
  <button class="ui button">Button</button>
</div>
```
[fiddle](http://jsfiddle.net/adcf0Ln5/2/)
Top of the button was not rounded


jlukic:
Thanks a lot for filing this issue, and including an easy to understand test case. I've added a milestone, and will try to take a look as soon as I can.

jlukic:
I've implemented this with slightly more succinct code.

Issue #4057, opened 2016-05-24
Disappearing text in multiple select within a popup


gdaunton:
I've found a bug where if a multiple select is placed within a popup, any entered text is not displayed in the input field until a few letters are typed.

I think it is something to do with the width being off with the `<input class="search" autocomplete="off" tabindex="0">` input that is within the dropdown.

I also had the same issue using the normal `<select>` version of the dropdown.

Also removing a tag causes the popup to close.

Here is my [jsfiddle](http://jsfiddle.net/ezgav7jy/) that should recreate the issue.

PR #4108, opened 2016-06-07
[Icon Input] Fix for issue #4057

gdaunton:
A fix for my own issue #4057 which pertained to having a multiple select dropdown inside a popup that was attached to an icon input.

The fix is that an icon input applies `padding-right: @iconMargin !important;` to all input children of an icon input. This caused problems with the inputs inside the popup which also received this padding. This also caused the search input text to disappear in the multiple select within the popup.


gdaunton:
just realize I added mdehoog's calendar stuff. Whoops...

PR #4136, opened 2016-06-18
[Menu] @dividerSize cannot be changed since it is hard coded

tobias-kuendig:
Removed duplicate height definition to respect the `@dividerSize` variable


jlukic:
Cheers

PR #4144, opened 2016-06-21
Fix compact form messages

### LINKED ISSUE #3343: Error Message `compact` class does not work
Demo: http://codepen.io/anon/pen/dYajjW

When an error message is given the `compact` class, it will not reduce it's size to fit it's contents. It still takes up the full width of the page. 

Removing the `error` class produces the intended results.

bcroq:
Add inline-block to compact form messages so they don't take the full width of their container.

Fixes #3343


jlukic:
Last one in!

PR #4157, opened 2016-06-22
[Icon Input] Fix for issue #4057

gdaunton:
A fix for my own issue #4057 which pertained to having a multiple select dropdown inside a popup that was attached to an icon input.

The fix is that an icon input applies padding-right: @iconMargin !important; to all input children of an icon input. This caused problems with the inputs inside the popup which also received this padding. This also caused the search input text to disappear in the multiple select within the popup.

### DIFF
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -273,7 +273,7 @@
 .ui.icon.input > i.icon:not(.link) {
   pointer-events: none;
 }
-.ui.icon.input input {
+.ui.icon.input > input {
   padding-right: @iconMargin !important;
 }
 


jlukic:
In `2.3.0` popup will now correctly calculate offsets with popup's in a different offset container, so you can now use `movePopup: false` in your example and prevent the css rules.

I'm also adding child selectors to all input fields to extend this PR.

seteam1:
This change restricts the ability to leverage a trending style when defining ui inputs. There is a growing trend towards nesting inputs within the label itself rather than as a sibling. The change forces input fields to be direct descendants of a ".ui.input" element.

Benefits include: 

- onclick events bubble across the entire label not just the input element, making it easier for users to interact with them through the label. Check boxes are a great example where they could click any part of the label and it will check/uncheck the box because of the nesting.
- vertical  alignment: It makes it far easier to vertically align the label with the input field without any custom css to center it.
- Nesting inputs within labels creates an implicit relationship for user clicks rendering the "for" & "id" properties unnecessary to make it so that clicking a label checks a box or gives a text field focus.


Is there any chance we can roll this back and consider addressing the issue in some other manner without imposing such a strong mandate about how to format a "ui input" element? This was a breaking change for our organization.
 

Thanks!

PR #4205, opened 2016-07-02
Add font-family to label inside an icon fixes #4204

### LINKED ISSUE #4204: [Label] font-family missing on label when used inside a icon
I created a floating label on an icon to create the following effect;
floating label inside icon

I noticed that the font-family is not applied to the label when used inside an icon.

brianveltman:


jlukic:
I would recommend wrapping the icon with a position relative container, I don't think the library should recommend nesting other components inside an icon.

PR #4206, opened 2016-07-02
[Icon] Path to 'Icons' font should not contain backslashes

diolive:
Replace path.sep with '/' char for build correctly under Windows (default path.sep == '\')

### DIFF
--- a/tasks/config/project/config.js
+++ b/tasks/config/project/config.js
@@ -92,9 +92,9 @@ module.exports = {
 
     config.paths.assets = {
       source       : '../../themes', // source asset path is always the same
-      uncompressed : '.' + path.sep + path.relative(config.paths.output.uncompressed, config.paths.output.themes).replace(/\\/g, '/'),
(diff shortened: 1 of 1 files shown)


brianveltman:
Good catch!
Same should be done for SUI base when installing. Now it changes over and over again when working on Windows and Unix since NPM rewrites semantic\\ to semantic// and vice versa. If semantic.json is not in your gitignore you'll be checking in these changes over and over again.

I think it should be somewhere around this:
https://github.com/diolive/Semantic-UI/blob/4f929b771e23fd7eec6f83a84e43a93f8bb410fd/tasks/config/project/install.js#L314

eXtreme:
:+1: on that, hit that problem when upgraded my build to 2.2 and now icon paths contain backslash.

aloker:
I'm having the same issue and the PR fixes it.

diolive:
It seems like author have already fixed this issue in https://github.com/Semantic-Org/Semantic-UI/commit/961b4c90680e8d3c1e423fbf9c0fdaf20d1e87af without merging my PR.

Issue #1336, opened 2014-11-30
[UI] Input Slider / Range Component


q-depot:
Hi,

does the UI comes with sliders? I couldn't find it anywhere in the website so I assume it doesn't, but I saw the buttons can be displayed as (toggle)sliders. That's not what I'm looking for but I thought there might be a widget somewhere.

Another note on the side, it would be cool to have a knob widget.

sevki:
:+1:

nicbarker:
Would love this feature, don't enjoy having to use the jquery UI component but it does the job at the moment.

hrstoyanov:
jlukich,
Where is the documentation on the new Range component. I could not find it at beta.semantic-ui.com

Thanks for your hard work!

jlukic:
There is not a new range component with 2.0. Not enough time.

mjhasbach:
+1

mrjonny2:
+1

rachellji:
+1

scotthasbrouck:
+1

michahell:
+1

pfhor:
+1

avalanche1:
+1

jeanlescure:
+1

pulsingbrain:
+1

Timonzimm:
+1

jlukic:
I started working on this

avalanche1:
Jack, please include:
- a vertical variation
- a way to add ticks with labels to slider axis

jlukic:
I plan on supporting both of those cases.

avalanche1:
Nice to have you back!)

mrjonny2:
Any word on when this will be done? Sorry to rush 😄

pashadia:
+1

vakorovin:
+1

lsunsi:
+1 Awesome work man (:

namwkim:
+1

miroslavbucek:
+1 thanks!

David-Desmaisons:
+1

kataras:
+1 Slider is useful, unfortunaly I am still using a plugin named 'bootstrap-slider', because I couldn't find any slider component for semantic ui and I am migrating all my sites from bootstrap to semantic ui.

tyleryasaka:
[...]

Issue #276, opened 2013-10-25
Site Navigation: Unable to scroll


ekanna:
Hi @jlukic 

Not able to scroll when cursor is pointed at code area. I don't have this problem with getbootstrap.com site. So it must be a bug.

Thanks

PR #4212, opened 2016-07-04
Range slider module

gdaunton:
Implemented and added onto the range component originally made by [tyleryasaka](https://github.com/tyleryasaka).

[His original repo](https://github.com/tyleryasaka/semantic-ui-range)

[Demo](https://jsfiddle.net/rjxb7ffy/)

[Issue](https://github.com/Semantic-Org/Semantic-UI/issues/1336)

I'm new to the whole javascript plugin thing, so let me know if I formatted something wrong or whatever.

Also let me know of any other features you would like to see!


bretto36:
Any chance you can update the indenting in the files so they are formatted a bit cleaner?

gdaunton:
Sure I'll give it a look. Any specific areas I should fix?

bretto36:
perfect. So much easier to read

chrisbell08:
Just has a look at the js fiddle for this... looks awesome!! Any idea when we can expect this to be released?

gdaunton:
That is up to @jlukic. 

Personally, I think this needs some more work until it is ready. There are a few features like step indicators and a vertical variation that I haven't had a chance to get to yet.

You can, however, use it in its current version if you like. Just include range.css and range.js in `dist/components` from the [`range-slider-build` branch of my fork](https://github.com/gdaunton/Semantic-UI/tree/range-slider-build) in your project.

chrisbell08:
@gdaunton sweet thanks!

gdaunton:
Added in a vertical variation. See update demo

gdaunton:
Added in support for keyboard events for the following keys:
  - increase
    - up arrow (depending on orientation)
    -  right arrow (depending on orientation)
    - page up
  - decrease
    - down arrow (depending on orientation)
    - left arrow (depending on orientation)
    - page down

Any other keys I should add?

Also looking into mouse wheel / scrolling support.

abijames:
Thanks for the work. I am working on a project to create an accessible website using Semantic-UI so really appreciate any improvement in keyboard accessibility. 

The only other action you could consider is adding Page Up and Page Down for Increment or decrement the slider by an amount larger than the step changes made by the arrow keys (as described at the [W3C design patterns for widgets](https://www.w3.org/TR/wai-aria-practices/#slider)). 

If you have a demo page with keyboard events I'd be happy to check it out with a screen reader to make sure its all working well.

gdaunton:
@abijames thanks for the suggestion! 

I made the page up/down changes and added a new setting which is called `pageMultiplier` which is the multiplier against the normal step that is used for the page keys. 

i.e. with a `step == 1` and `pageMultiplier == 2`, the slider will increase by a value of 2 when page up is pressed. 

Also I am not familiar with how tabbing is supposed to work, but I added the `tabindex` attr so that the slider component can be focused on.

Also this commit should fix the issues with the jsfiddle 😄.

jlukic:
I'm interested in reviewing this, but I think it might be missing what is the most common use of a slider--being able to set a range with a lower and upper bound.

gdaunton:
@jlukic OK I'll take a look at adding that this weekend 👍

gdaunton:
@jlukic Added in upper and lower bound, see updated demo.

nblackburn:
@gdaunton Nice work, on staged ranges, it would be better if it were to lock/jump to each stage instead of being free flowing though.

gdaunton:
@nblackburn hmm, it should snap when you stop dragging.

nblackburn:
@gdaunton Yes, that's right :+1:

jlukic:
Nice work, I agree it should probably snap before dragging completes so that its clear the final value of the range without having to trigger `mouseup` to verify.

gdaunton:
Should I make it an option to do the smooth scrolling, or just make the snapping the default behavior?

nblackburn:
@gdaunton Default if there is ranges defined, smooth otherwise.

gdaunton:
OK changed. Added a new `smooth` setting for the old behavior just in case people want it.

jlukic:
I have to say this is one of the most complete and consistent UI contribution the project has received yet. Everything looks good, just need to review and create docs.

One thing, noticing custom labeled range is behaving a bit strangely, resting position appears to miss some of the labels.

gdaunton:
@jlukic Thanks! I was just about to ask for some more suggestions.

Yeah I was torn on the custom label behavior. At the moment if it doesn't have a `data-value` attr, they are spaced evenly across the slider.

What do you mean by resetting position?

gdaunton:
Oh resting position, ha never mind.

nblackburn:
@gdaunton Tiny thing, you shouldn't be able to hover over a disabled range handle (currently triggers a transition).

gdaunton:
@nblackburn fixed. Demo updated.

@jlukic thoughts on the custom label behavior? Should it be that each label is put at a step sequentially (i.e. the first label element is put at the first step value)?

nblackburn:
@gdaunton Nice, hopefully this will be merged in soon :)

etasbyjh:
@gdaunton Do you know why this label thing goes wiered? it also affects on the 'start' value... actual value range from the range slider is in a correct range though.
Thanks!

image

gdaunton:
Could you put your code in a jsfiddle? Here's one with the resources already loaded: https://jsfiddle.net/reoLaufj/.

etasbyjh:
@gdaunton Hello! Thanks for the quick response :) 

Please check my code from the link below.

https://jsfiddle.net/reoLaufj/1/

gdaunton:
@etasbyjh All fixed your fiddle reflects the changes. SO MANY LABELS

etasbyjh:
@gdaunton  Thanks for the FIX! 

Can I ask one more thing?... I am still new to github system... which file have you updated??... I mean where can I get the update range.js and range.css files? Thanks!

gdaunton:
Yeah no problem. Wrote a previous comment about this [here](https://github.com/Semantic-Org/Semantic-UI/pull/4212#issuecomment-231544947).

You can download the branch as a zip, extract it, and use the files from the comment.

etasbyjh:
@gdaunton Sorry... I may have found one more issue. 

Setting min / max value has some issue. Could you please check this one as well?
https://jsfiddle.net/reoLaufj/3/

Thank you very much!

gdaunton:
@etasbyjh no problem! Keep the bugs coming!

I'll give it a look.

etasbyjh:
@gdaunton May I ask one more thing? Would it be possible for you to make a change (or add a new function) to be able to get a value instantly?(while user scrolling rather than after scrolling) With the original version(by tyleryasaka), I could get a value during scrolling by onChange function. 

Thank you so much!

gdaunton:
@etasbyjh  sure. There is actually already an `onMove` callback that does exactly what you described.

etasbyjh:
@gdaunton wooops! of course ;p  Thanks for the tip. 

I have to say I do appreciate all of your helps.

gdaunton:
@etasbyjh No problem at all!

etasbyjh:
@gdaunton  Hi! I brought a new bug... D:  now with android tablet... I guess touch event has some issues with range component. Please see the attached! Thanks!

image

gdaunton:
@etasbyjh Awesome! I pushed a fix that will hopefully work. Let me know if it does.

etasbyjh:
@gdaunton Always appreciate your super quick fixes!!  but D: still there is an issue... it definitely moves while dragging but cannot exit from its dragging event. so after entering click > dragging events, it never releases the dragging event. meaning.... cannot set a value.

hope this makes sense...

Super Thanks!

gdaunton:
@etasbyjh Yep I just realized that bug myself. Taking a look!

gdaunton:
@etasbyjh OK it seems to be fixed on my end. Can you confirm?

etasbyjh:
@gdaunton  WORKING!!! cannot thank you enough!!!!

jlukic:
I've begun reviewing and making modifications. The semantic ui docs repo will also need to be updated to include the definition, usage, examples, settings etc for this component which will take a bit of time along with the review

gdaunton:
@jlukic OK I'll freeze any updates. Let me know if you want any help with the docs.

etasbyjh:
@gdaunton @jlukic Thanks both!  I look forward to it.

mbaileys:
Thanks for the great work on range sliders! Would it be possible to grab min/max/step values and other attributes from the DOM so that $('.ui.slider').slider() works? Since they're already defined for <input type="range"> it would make conversion simpler as well. Instead of giving onChange explicitly, we could use a <input type="hidden"> like dropdowns do. This would help make the slider usable in a virtual-dom app when multiple dynamically generated sliders can coexist.

gdaunton:
@mbaileys yes you can already do this as you can with other semantic ui elements. You'll be able to see all of the methods you can use once the documentation is completed.

I would defiantly advise waiting until we get the slider finalized and in the master codebase before you use it in the real world. Thanks!

artsocode:
Hi. Thanks for the semantic-ui! And special thanks for add range control! But i guess, i found the bug! Range incorrect show **start** position in **modal**. Take a look.

https://jsfiddle.net/rjxb7ffy/2/

gdaunton:
@jlukic What do you think about this bug? 

The bad positioning is because of the slider width change, which I solved previously with the accordion bug. But, with this case, I don't think it will work since the slider doesn't get an attribute change like it does with the accordion. 

Thoughts on a fix?

artsocode:
@gdaunton Probably i have no idea, how to fix that. But if i use percent for **track-fil** and **thumb** in modal, he work fine! Maybe calc his start width in percent and next use px for calc steps and etc.

For example.
If we have **min: 0**, **max: 300**, **start: 15**
`start_width = 100 * start / max; (5%);`

Maybe i say stupid things... 
but think about it. Thanks!

Sorry for my English :(

gdaunton:
@artsocode Hey no problem! Your english is fairly good!

Yeah having a static width (%, px or em) for the slider would fix the problem since the bug results from how the slider changes width between the modal initializing and opening. However I think the standard behavior for the slider should be to fill the container it is in.

Not a bad thought though!

flytreeleft:
@gdaunton I found there are two issues should be fixed in `range.js`.

First, observing all attributes' mutation will cause an endless `callback` invoking when resizing chrome devtools panel which is docked at bottom of current window.

gtkshot

It seems that [Sizzle](https://github.com/jquery/sizzle) will set a temporary `id` to html node when resizing window, after a short time, it will set `id` to `null`. So, `module.event.resize` will be invoked endlessly.

I think only `css` and `style` needs to be observed, they will change the size of node. There is the code:
```diff
--- a/semantic/src/definitions/modules/range.js
+++ b/semantic/src/definitions/modules/range.js
@@ -137,6 +141,8 @@ $.fn.range = function(parameters) {
           slider: function() {
             sliderObserver.observe($module[0], {
               attributes: true,
+              attributeOldValue: true,
+              attributeFilter: ['class', 'style']
             });
           },
         },
```

It maybe better if we can [observe CSS style changes](http://xml3d.org/xml3d/specification/styleobserver/).

Second, the label position should plus the padding of the range container. There is my solution: 
```diff
--- a/semantic/src/definitions/modules/range.js
+++ b/semantic/src/definitions/modules/range.js
@@ -192,6 +192,7 @@ $.fn.range = function(parameters) {
               $children   = $labels.find('.label'),
               numChildren = $children.length,
               ratio,
+              margin = module.get.trackMargin(),
               position
             ;
             $children.each(function(index) {
@@ -212,27 +213,29 @@ $.fn.range = function(parameters) {
                 :
                 module.is.reversed() ? 'right' : 'left'
               ;
-              $(this).css(posDir, position);
+              $(this).css(posDir, position + margin);
             });
           },
           autoLabel: function() {
             if(module.get.step() != 0) {
               $labels = $module.find('.labels');
               if($labels.length != 0)
-                $labels.empty()
+                $labels.empty();
               else
                 $labels = $module.append('<ul class="auto labels"></ul>').find('.labels');
+              var margin = module.get.trackMargin();
               for(var i = 0; i <= module.get.numLabels(); i++) {
                 var
                   $label = $('<li class="label">' + module.get.label(i+1) + '</li>'),
-                  position =
+                  position = module.determine.positionFromValue((i+1) * module.get.step() + module.get.min()),
+                  posDir =
                     module.is.vertical()
                     ?
                     module.is.reversed() ? 'bottom' : 'top'
                     :
                     module.is.reversed() ? 'right' : 'left'
                 ;
-                $label.css(position, module.determine.positionFromValue((i+1) * module.get.step() + module.get.min()));
+                $label.css(posDir, position + margin);
                 $labels.append($label);
               }
             }
@@ -484,6 +487,15 @@ $.fn.range = function(parameters) {
           trackEndPos: function() {
             return module.is.reversed() ? module.get.trackLeft() : module.get.trackLeft() + module.get.trackLength();
           },
+          trackMargin: function () {
+            var modulePadding = 0;
+            if (module.is.vertical()) {
+              modulePadding = module.is.reversed() ? $module.css('padding-bottom') : $module.css('padding-top');
+            } else {
+              modulePadding = module.is.reversed() ? $module.css('padding-right') : $module.css('padding-left');
+            }
+            return parseFloat(modulePadding);
+          },
           precision: function() {
             var
               decimalPlaces,
--- a/semantic/src/definitions/modules/range.less
+++ b/semantic/src/definitions/modules/range.less
@@ -169,10 +169,12 @@
 
 .ui.labeled.range > .labels {
   height: @labelHeight;
-  width: 100%;
+  width: auto;
   margin: 0;
   padding: 0;
   position: absolute;
+  right: 0;
+  left: 0;
   top: 50%;
 }
 
@@ -200,9 +202,10 @@
 
 .ui.labeled.vertical.range > .labels {
   width: @labelHeight;
-  height: 100%;
+  height: auto;
   left: 50%;
   top: 0;
+  bottom: 0;
   transform: translateX(-50%);
 }
```

Finally, thanks for your effort 👍!

Banandrew:
Hey, @gdaunton

There’s common a problem with jQuery 3+ where code written for older versions produces this error: `Uncaught TypeError: elem.getClientRects is not a function`—or something along those lines. For example, it can break SUI popups, please see [this fiddle](https://jsfiddle.net/Banandrew/15oo8fxw/) with the developer console opened (it’s a fork of the range module’s demo). The reason I’m posting it here, is because SUI became compatible with jQuery 3 since 2.2.0, popups work fine with jQuery 3+, and I didn’t find any issues tied to this error; can it be a problem with the range module?

This error is easily fixable by plugging in [jquery/jquery-migrate](https://github.com/jquery/jquery-migrate), but it should only be a temporary solution.

Banandrew:
@gdaunton 

I played with it some more, and noticed a problematic behaviour: look at [this gif](http://i.imgur.com/B3Sm5Fx.gifv), if I drag and release the pin somewhere out of the range’s container, it doesn’t set the value, and I have to be very precise to set it. The step value in this gif is 500, so the lower the step, and the higher the min/max values compared to this step, the more difficult it is for the user to set the range’s min/max values. For example, if I set it to 0 instead of 500, it’s going to be very difficult to set min/max values.

I think it’s a big UX issue that should also be addressed in the future before the merging, as probably all of the popular standalone plugins don’t have it. A good example is [Ion.RangeSlider](http://ionden.com/a/plugins/ion.rangeSlider/demo.html).

gdaunton:
Thanks to you both for taking a look!

@flytreeleft 

These changes look great! I would recommend you open a pr on this repo against the `slider` branch. Would really like to check your code out myself. Also that way if we want to talk about any other changes we can do so without adding more comments to this pr.

@Banandrew 

I've found the jQuery issue. It appears that it must be related to [the `offset()` changes](https://jquery.com/upgrade-guide/3.0/#breaking-change-invalid-input-to-the-offset-method). I am guessing something about [this line](https://github.com/Semantic-Org/Semantic-UI/pull/4212/files#diff-0a3b03a7a9ced3ca79dafdd4a521f1c2R462) is causing the error. I'll dig a bit deeper when I get home.

Huh, I did not notice that behavior and it is defiantly not intentional. I will give that a look as well.

Again, thank you both!

flytreeleft:
@gdaunton OK, I will create a PR as soon as possible.

jlukic:
Just want to make it clear I do really want to get this into SUI. Apologies for any delays.

gdaunton:
@Banandrew ok a bit of an update on the jQuery issue. Turns out that this is just a git issue. Since the move to 3.0 happened a bit after (or during? I have no idea) my PR, my fork is behind on all of those changes. 

So, because of that, the popup will not work correctly **on my fork** with 3.0.

 That being said once everything is merged and finished, that will no longer be an issue.

**TL;DR:**  ¯\_(ツ)_/¯ It'll fix itself.

src/definitions/modules/range.js
```diff
@@ -362,6 +370,10 @@ $.fn.range = function(parameters) {
             ;
             if(eventPos >= module.get.trackOffset() && eventPos <= module.get.trackOffset() + module.get.trackLength()) {
               module.set.value(module.determine.value(newPos));
+            } else if(eventPos >= module.get.trackOffset()) {
+              module.set.value(module.get.max());
+            } else if(eventPos <= module.get.trackOffset() + module.get.trackLength()) {
```
flytreeleft: `<=` --> `>`

src/definitions/modules/range.js
```diff
@@ -362,6 +370,10 @@ $.fn.range = function(parameters) {
             ;
             if(eventPos >= module.get.trackOffset() && eventPos <= module.get.trackOffset() + module.get.trackLength()) {
               module.set.value(module.determine.value(newPos));
+            } else if(eventPos >= module.get.trackOffset()) {
```
flytreeleft: `>=` -> `<`

src/definitions/modules/range.js
```diff
@@ -352,6 +352,14 @@ $.fn.range = function(parameters) {
                   settings.onMove.call(element, value);
                 });
               }
+            } else if(eventPos >= module.get.trackOffset()) {
```
flytreeleft: `>=` -> `<`

src/definitions/modules/range.js
```diff
@@ -352,6 +352,14 @@ $.fn.range = function(parameters) {
...
+              module.update.position(
+                module.determine.positionFromValue(module.get.max())
+              );
+            } else if(eventPos <= module.get.trackOffset() + module.get.trackLength()) {
```
flytreeleft: `<=` -> `>`

flytreeleft:
@gdaunton I have pull a request to `slider` branch, you can check the changes in #4678.

Urahara:
@gdaunton First thank you for all this work, is there any chance to add this behaviour to the component? http://blog.153.io/Elessar/ (I'm coding a app that user set that hours available for work, so i need multiple points in slider).

gdaunton:
@Urahara Hey, thanks for the suggestion!

That is an interesting use case! Unfortunately, ever since I traded my days for a paycheck, I have a lot less time to dedicate to programming for fun 😢 . As this is the case, I am only working on bug fixes for now.

I would suggest you checkout [needtomeet](http://www.needtomeet.com/schedule#/step2). They have a really great solution to the same problem.

If you don't find that helpful, you could try to re-skin that lib with your own `css` to be more semanticy. The existing [progress bar](http://semantic-ui.com/modules/progress.html) looks kinda similar!

joaumg:
Any update on this ? Need help to build a docs ?

gdaunton:
Started on docs here: https://github.com/Semantic-Org/Semantic-UI-Docs/pull/276

themre:
is this still depending on jquery? because latest release remove jquery dependency.

noinkling:
I hope this gets merged at some point. A feature I would really like to see is an option to have the background color originate at the middle (or even better, a specified start point) rather than the beginning of the slider.

gotenxds:
Any reason this is still not inside ?

jcttrll:
There are a couple of spots where `Object.assign` is used, which breaks in IE11 (11.0.9600.18697, if that's actually helpful). Could be worked around based on the polyfill example here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/themes/default/modules/range.variables
```diff
@@ -0,0 +1,80 @@
...
+/* Thumb Hover */
+@thumbHoverPointer    : pointer;
+@thumbHoverBackground : @whiteHover @subtleGradient;
+
+/*-------------------
+        States
+--------------------*/
+
+/* Disabled */
+@disabledOpactiy        : .5;
```
jcttrll: Little typo, here

jcttrll:
If the `onChange` callback throws an exception (which we all agree it shouldn't, but... :smiley:), the slider is "stuck" to the mouse pointer--even though the mouse button was released, the slider still follows the pointer. See: https://jsfiddle.net/g55dybda/

jesuGMZ:
Is not merged yet cause of the docs conflict #276?
Please, consider a final push to integrate this great contribution.

willstott101:
Possible bug: In the demo if you click and drag a reversed slider, and move your mouse off the side of the containing segment, the slider thumb snaps to the wrong end of the slider. Behaviour works fine on non-reversed sliders. Same problem on the reversed vertical and horizontal slider is affected.

Awesome work, would love to see this added!

PapsOu:
@jlukic : Does this PR is planned to be merged soon ?

Ashpabb:
Why is this closed and not merged?

noinkling:
This was already merged into the `slider` branch in this repo, with some additional fixes, however it's just sat there since then. I'm currently using it in a project, but in a painful way where I have to manually generate only the slider CSS using that branch and hope there aren't any incompatibilities with more recent versions of SUI. I don't know why it hasn't been merged into master.

#1336 is still open and one of the most popular issues, for whatever it's worth.

gdaunton:
I recently started using GitHub again for a project and saw this in my open prs. I closed it since it's been open for way too long.

xFirez:
Will this be added in?

gdaunton:
🤷‍♂️

Ashpabb:
Malcolm, thanks for the synopsis. Currently working on a port of this for
Semantic UI React

gdaunton:
@Ashpabb you might have trouble there since the react port gets all it's styles from this repo. As long as this component and it's styles are not included in core, your port will be in limbo.

Ashpabb:
Gareth, understood - nothing wrong with getting work done ahead of time ;)

lubber-de:
The slider module has been merged into https://github.com/fomantic/Fomantic-UI since 2.5 and empowered as of 2.7.0
See https://fomantic-ui.com/modules/slider.html

PR #4225, opened 2016-07-06
:memo: unreachable code detected

basarat:
Detected by running this project through TypeScript using [alm.tools](http://alm.tools/) :rose:

I suspect there is more wrong with this function as this unreachable code doesn't seem intentional, but I leave that to you :rose:


jlukic:
Should be removed from src, but yes is unreachable.

PR #4244, opened 2016-07-08
[Icons] Removed duplicated 'talk' icon with incorrect content

anantoghosh:
` i.icon.talk:before { content: "\f27a"; }` was re-declared in "Aliases" section with different content value `"\f075"`.
The alias was overriding the correctly declared talk icon.
You can look in the docs http://semantic-ui.com/elements/icon.html, the "Talk" icon should be same as "Talk Outline" (with just the outline of course).


jlukic:
Thanks for the PR, keep in mind this PR should be in `next`

anantoghosh:
Okay, thanks.

PR #4273, opened 2016-07-12
[Modal] Add ability to align to top of window.

### LINKED ISSUE #4272: [Modal] Add ability to align Modal to top of window
It would be useful to have the ability to show a modal attached to the top of the window.  This may be as simple as adding an `align` property to the Modal `settings` object, and if set to `top`, the Modal `set.position()` method sets `top` to 0.  This may be a naive implementation - I haven't done enough testing with scrolling pages or long modal boxes.

daggmano:
Added a Modal setting ‘align’ which may be ‘top’ or ‘middle’ (or
anything other than ‘top’).

Fixes #4272

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -623,22 +623,33 @@ $.fn.modal = function(parameters) {
             }
           },
           position: function() {
-            module.verbose('Centering modal on page', module.cache);
-            if(module.can.fit()) {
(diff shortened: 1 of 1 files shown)


jlukic:
I was just about to code this.. will review

jlukic:
I've implemented this using flexbox in `2.3.0`

PR #4311, opened 2016-07-20
[Dropdown] Disconnect.selectObserver bugfix

Paklausk:
Dropdown function 'disconnect.selectObserver' trying to disconnect menuObserver instead.


jlukic:
This is a bad typo.

PR #4316, opened 2016-07-21
[Dropdown] Disconnect.selectObserver bugfix

Paklausk:
Dropdown function 'disconnect.selectObserver' trying to disconnect menuObserver instead. #4311


jlukic:
Thank you

PR #4323, opened 2016-07-23
Fix for ‘get values’ returning strings for number fields

ubald:


jlukic:
Returning strings for all JS types is expected behavior. This is to keep parity with what can be stored in an `input`'s `value` and submitted with `form`

Issue #4336, opened 2016-07-26
Labels of disabled radio fields are not greyed out


louwers:
Unlike other fields that are `disabled`, the labels of radio fields are not greyed out when they are `disabled`. Screenshot from my project to demonstrate what I mean (all fields are set to `disabled`):

![screen shot 2016-07-26 at 14 13 28](https://cloud.githubusercontent.com/assets/649392/17137384/3452fdc8-533b-11e6-96cd-a938822ef39f.png)

Jsfiddle: http://jsfiddle.net/Lcj08k44/

jlukic:
Fixed in 0f6d1617ac4be4276695e51d7995d11e06adcc22

PR #4338, opened 2016-07-26
[Label] Label is not disabled with radio buttons disabled

GDay:
We only need to take the first label, otherwise it also affects the labels of the radio buttons, which we don't want.

Answer for: https://github.com/Semantic-Org/Semantic-UI/issues/4336
http://jsfiddle.net/y4vn6hgz/


jlukic:
I've implemented this in `2.2.3`

PR #4359, opened 2016-07-29
Use semantic-ui table with Jqgrid would be awesome

xtrepo:
I think to have a the best table is to use Jqgrid with perform Semantic-ui template 
What do you think ?


philrykoff:
DataTables is also very powerful.

jlukic:
This is not a PR

Issue #4275, opened 2016-07-12
[Dropdown] Multiselect Without Labels Does Not Clear


sangkj:
When using a multiselect dropdown (with useLabel:false option), the dropdown items do not clear properly.  When "get value" is called on the dropdown after "clear" is called, the dropdown returns an empty string (expected), but the selected menu items are not deselected in the ui (problem).  Here is a jsfiddle to demonstrate the issue http://jsfiddle.net/k8z7aqro/1/

jlukic:
I'm able to reproduce

PR #4366, opened 2016-07-30
Fix issue #4275

vinh123456789:
Fix issue #4275, when using useLabels: false, dropdown("clear") won't
remove class "active" from "item"

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2075,13 +2075,12 @@ $.fn.dropdown = function(parameters) {
 
         clear: function() {
           if(module.is.multiple()) {
-            module.remove.labels();
-            module.remove.activeItem();
+            module.remove.labels();            
           }
-          else {
-            module.remove.activeItem();
+          else {            
             module.remove.selectedItem();
           }
+          module.remove.activeItem();
           module.set.placeholderText();
           module.clearValue();
         },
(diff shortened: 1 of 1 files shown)


davidkuhta:
If we're calling `module.remove.activeItem();` inside both the `if` and the `else`, shouldn't it be brought outside?
Something like:
```
clear: function() {
  if(module.is.multiple()) {
    module.remove.labels();
  }
  else {
    module.remove.selectedItem();
  }
  module.remove.activeItem();
  module.set.placeholderText();
  module.clearValue();
},
```

vinh123456789:
You are right...
This is my first time contributions to a github's project, i'm so nervous that i will make a mistake, and here is it...

davidkuhta:
Well technically it wasn't a mistake because your code worked, we just refactored. 😄 
And whatever way @jlukic goes with this PR, thanks for helping to make Semantic UI better.

jlukic:
`remove labels` calls `remove values` so this would cause values to remove twice with labels.

Probably need
```javascript
clear: function() {
  if(module.is.multiple() && settings.useLabels) {
    module.remove.labels();
  }
  else {
    module.remove.selectedItem();
  }
  module.remove.activeItem();
  module.set.placeholderText();
  module.clearValue();
}
```

jlukic:
Confirmed fix

vinh123456789:
@jlukic is there any reason why `module.remove.activeItem()` can't be place outside of the `if else` statement?

jlukic:
It is placed outside?

Issue #4398, opened 2016-08-05
Icon group should accept `link` class


tbracken:
Fiddle: http://jsfiddle.net/g46jhxpy/2/

Icon groups should be able to use `.link` to style as clickable.

IonutBajescu:
Closing the issue since the PR has been merged.

PR #4399, opened 2016-08-05
Added hover styling to icon group

tbracken:
issue #4398 (Working screenshot below, hovering on the right)
screen shot 2016-08-05 at 4 50 24 pm


jlukic:
Thanks!

PR #4412, opened 2016-08-09
Added Material Design Icons to Semantic UI material theme.

nielslerche:


jlukic:
Merged in 49444b6a101a25f28ac9e4117bcf3c5168c9c25b I've modified paths to be more consistent with rest of lib

nickpoulos:
It looks like the actual asset names don't match the default icon.variables. 

I just cloned the repo and tried it. The default path is: 

@fontPath: './../themes/innohead/assets/fonts';

it should be:

@fontPath: '../../themes/material/assets/fonts';

Then also the font name doesn't match. The binary asset files are all named "icons.eot", "icons.woff", etc. They should be named "MaterialIcons-Regular.eot", "MaterialIcons-Regular.woff", etc.  

OR

The @fontName should be set to: 

@fontName: 'icons';

Unless I am missing something?

nickpoulos:
This also breaks Semantic UI standard styles for icons inside input.

.icon.input:before { content: '\e890'}

How/Why did this get merged?

3zzy:
Awesome! How do I use material icons instead of FontAwesome? The documentation doesn't say anything about it.

nielslerche:
If you have the material theme for Semantic UI installed, you should be able to change the `@icon` property value to `material` in the theme config.

neekey:
Hi, I noticed that there is not a `.woff2` font file included in this theme, according to the source code for Icon https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/elements/icon.variables#L13, 5 type of font files should always be presented, and this will cause build error when I tried to build it with webpack

Issue #4413, opened 2016-08-09
[Validation] - should not check elements that have display: none css


avalanche1:
https://jsfiddle.net/avalanche1/rgacmkzL/2/
Name input should not be validated.
This is required when form has dynamic field selection, eg field selection changes based on the 1st field value.

avalanche1:
https://github.com/Semantic-Org/Semantic-UI/pull/4416

PR #4416, opened 2016-08-09
[Validation] - should not check elements that have display: none css 

avalanche1:
check if field has display: none
https://github.com/Semantic-Org/Semantic-UI/issues/4413


jlukic:
This might be complicated as we currently support validating hidden fields.

Issue #4422, opened 2016-08-10
[Validation] - 'on: change' causes error prompts to appear when blurring field


avalanche1:
https://jsfiddle.net/avalanche1/f7mbax9q/ - click name field and then anywhere else  - error prompt appears.
I would not expect that behaviour from this setting. I would expect validation to run only on `input` event.

PR #4423, opened 2016-08-10
'on: change' causes error prompts to appear when blurring field 

avalanche1:
https://github.com/Semantic-Org/Semantic-UI/issues/4422


jlukic:
Accidentally merged into master, manually merged into `next`.

PR #4431, opened 2016-08-12
Update site.variables

morenoh149:
fix twitter brand color


jlukic:
This is already fixed in `next` release
https://github.com/Semantic-Org/Semantic-UI/blob/next/src/themes/default/globals/site.variables#L428

morenoh149:
cc @vivekbhagwat

Issue #4501, opened 2016-08-29
[Horizontal List] Strange alignment in smaller screens


renatodinhani:
When displaying the horizontal list in a small screen, the wrapped items have this strange space in the left side instead of perfectly aligning like a grid.

![horizontal list](https://cloud.githubusercontent.com/assets/1139781/18065224/61ca649c-6e09-11e6-856b-3e9caefbdda4.png)

This is happening because the horizontal list is using margin-left to separate the elements (and setting the margin of the first element to zero). I believe a better option is to use margin-right and set the margin of the last element to zero.

LeonardoGentile:
This still seems to be a problem. 
This also happens in large screens when we have a large number of items, when they span on the next row then the issue is still visible

lubber-de:
The original PR's #4701 and #4502 both forgot to take care of the `divided horizontal` variant.
We finally also took care of it :slightly_smiling_face: See https://github.com/fomantic/Fomantic-UI/pull/353

PR #4502, opened 2016-08-29
Fix for #4501

dinhani:
This PR makes the horizontal list to use margin-right instead of margin-left to keep the left alignment when the list items wraps due to smaller screens.


Banandrew:
Linking the issue here: #4501.

Issue #4183, opened 2016-06-28
Single dropdown selection does not work when clicking an item and having 'select' action (version 2.2.1)


dorsha:
Installed version: 2.2.1

I have the following dropdown configuration:

```
      dropdownSelector
          .dropdown({
            fullTextSearch: true,
            preserveHTML: true,
            onChange,
            action: 'select',
            onHide: this.onHide,
            onShow: this.onShow,
            transition
          });
```

I am using the 'select' action since I don't want the current text to be changed.

Prior version 2.2.x, it works fine.
After the upgrade, the 'onChange' is not dispatched when clicking the dropdown item - but, it dispatched when moving around with the keyboard arrow keys.

I also tried to add `allowReselection: true` but it did not fix it.

Any idea?

Clicking the item is not doing anything, but moving around with the keyboard works.
![image](https://cloud.githubusercontent.com/assets/5429033/16410042/3e3cfe06-3d29-11e6-9d43-18d7c9f35a92.png)

When I remove the 'select' action it works, but it changed the selected text.

jlukic:
Can you create a JSFiddle? There were significant dropdown changes and I want to make sure everything is fine.

dorsha:
Yes.

**Semantic-UI 2.1.8**:
https://jsfiddle.net/3kcftobu/
Clicking on the dropdown's items works fine (you get an alert).

**Semantic-UI 2.2.1**:
https://jsfiddle.net/y2vq2u8m/
Clicking on the dropdown's items does not trigger the alert.
It works when you try to navigate with the keyboard arrows keys. (in the previous version, navigate with the keyboard keys was not triggered the onChange event)

Ignore the bad style :)

yaowenz:
I've got the same issue on 2.2.1

assafg:
Any updates on this issue?

Kukkimonsuta:
The 'onChange' trigger on keyboard arrows keys seem to be intended change and you can disable it by setting new option `selectOnKeydown` to `false`. This however still doesn't change the fact, that `onChange` doesn't trigger on either `click` or `enter` when `action == 'select'`.

**Semantic-UI 2.2.1**
https://jsfiddle.net/y2vq2u8m/1/

dorsha:
@jlukic, any update regarding this issue?
It's blocking us for upgrading to Semantic-UI 2.2.x

jlukic:
Working on this now.

jlukic:
[...]

PR #4510, opened 2016-08-31
Fix for #4183

martinduparc:
`text` argument missing for select action


jlukic:
I've read through the original thread #4183 and it is still an issue. I tested this fix and it looks good.

PR #4517, opened 2016-09-01
[Popup] support svg

Tweety79rw:
Originally set to === which would always return false if the browser supports svg graphics element


Banandrew:
Hi @Tweety79rw, thanks for posting in the thread, linking it here: #5360.

lubber-de:
Fixed since 2.6.4 https://github.com/fomantic/Fomantic-UI/pull/247

PR #4544, opened 2016-09-10
Correct check for undefined variable

mathiasrw:
The code was comparing the result of `typeof` with a variable named `undefined`.

As typeof returns a string it should compare to `'undefined'`

---

This change is [Reviewable](https://reviewable.io/reviews/semantic-org/semantic-ui/4544)


jlukic:
Should be `==` or `undefined` that is indeed a mistake

mathiasrw:
:+1:

Issue #4497, opened 2016-08-29
[Form Validation] JS error when revalidating inputs without defined validation rules


cbxp:
In our tests, we often get this error:
`Uncaught TypeError: Cannot create property 'identifier' on boolean 'false'`

From this code:

```
field: function(field, fieldName) {
...
            if(!field.identifier) {
              module.debug('Using field name as identifier', identifier);
              field.identifier = identifier;
            }
```

This happens when there are no validation rules defined and function argument `field` gets the value `false`. Most definitely, this condition should be checked.

```
          change: function(event) {
              var
                $field      = $(this),
                $fieldGroup = $field.closest($group),
                validationRules = module.get.validation($field) // FALSE RETURNED FROM HERE
              ;
              if(settings.on == 'change' || ( $fieldGroup.hasClass(className.error) && settings.revalidate) ) {
                clearTimeout(module.timer);
                module.timer = setTimeout(function() {
                  module.debug('Revalidating field', $field,  module.get.validation($field));
                  module.validate.field( validationRules ); // FALSE PASSED HERE
                }, settings.delay);
              }
            }
```

wenweih:
#4221

jlukic:
There should be a check

wenweih:
@jlukic should fix the validate error when there is search dropdown in form.

cbxp:
We have a simple fix: https://github.com/Semantic-Org/Semantic-UI/pull/4547

PR #4547, opened 2016-09-12
#4497 check if validation rule is defined when revalidating

cbxp:


jlukic:
I've manually merged this into next, fix looks good. 

Thanks!

Issue #1802, opened 2015-02-13
inverted flat tcheme + input, text are invisible


agborkowski:
![screen shot 2015-02-13 at 11 53 35](https://cloud.githubusercontent.com/assets/170557/6186227/16368f86-b377-11e4-9b13-2ce5cab292a3.png)

#source http://semantic-ui.com/collections/form.html

Anachron:
Not on my iPhone, which OS/browser do you use?

jlukic:
Themes other than the `default` theme are not supported. If you are interested in improving or hacking upon them PR will be accepted.

agborkowski:
disappears on focus and text is written in black so is invisible, Chrome Version 40.0.2214.115 (64-bit), ill try fix it.

PR #4550, opened 2016-09-13
Fix fg-color on inverted forms in flat theme.

musically-ut:
This is an attempted fix for #1802, which was closed as a `Cannot fix`.

# Original version

On the current website, with the Flat theme selected.

# Fixed version


jlukic:
Thank you

PR #4430, opened 2016-08-12
FIX: auto install did not copy build during npm install

AnsonT:
With 'autoInstall: true' in semantic.json, npm install does not copy out the semantic files to root directory. 

The following check always fails when install.shouldAutoInstall() === true

``` javascript
 /*--------------
    NPM Install
  ---------------*/

  // Check if PM install
  if(answers.useRoot || answers.customRoot) {
```

PR #4567, opened 2016-09-19
Feature/fix autoinstall

AnsonT:
Semantic.json
- "autoinstall" : true - will now copy to specified semantic path
- "build": true - will run gulp build on the copied path


jlukic:
I dont understand the logic here. If we're building for continuous integration (CI) or other autoinstall we'll need the files to be built

AnsonT:
These 3 files are missing the additional commits. Let me create a new pull request.

AnsonT:
The main fix for autoinstall was merged https://github.com/Semantic-Org/Semantic-UI/pull/4430. This was intended to give the user an option top copy and not build on autoinstall, not needed.

PR #4574, opened 2016-09-21
Fix broken icon headers with content icons

philrykoff:
Well, as the title says. Maybe renaming the icon would be a better idea, as it may interfere in other areas than the header module.

### DIFF
--- a/src/definitions/elements/header.less
+++ b/src/definitions/elements/header.less
@@ -96,20 +96,20 @@
      Content
 ---------------*/
 
-.ui.header .content {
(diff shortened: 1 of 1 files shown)


jlukic:
`Content` includes two aliases `sidebar` and `bars`. 

Content was definitely a bad name choice. I'll change the docs to use the more compatible `sidebar icon`, as this is the most common use for the triple bars.

jlukic:
Changed in docs in next release. Thanks for input!

PR #4583, opened 2016-09-23
Stop adding the 'success' class to a form when data validation is successful

### LINKED ISSUE #3683: [Form] Form ALWAYS get the "success" class?
I'm sending the form over ajax capturing the "submit" event `$("#form").on("submit"...)` using `e.preventDefault();` to stop the POST default send (which is still configured as a fallback, just in case) and, as a callback to the ajax call, setting the form state as "success" or "error"... BUT no matter what the response is, the form always gets the "success" class.... why? Is it just because it passed validation? Because then, the [example](http://semantic-ui.com/collections/form.html#success) on the page is misleading as hell 

> Form Completed
> 
> You're all signed up for the newsletter.

meaning: "THE SERVER RECEIVED AND UNDERSTOOD YOUR PETITION", which you can't be sure of until you receive the server's response, so it should be activated programatically. So... is there anyway to override this (seemingly useless) behaviour? I've searched the docs and I didn't find anything...

chris-griffin:
I'm just deleting the two lines that add the 'success' class to a form when data validation success occurs. Data validation success is usually communicated to the user via a lack of error messages, not via a success message. A success message should only show within the form when it was successfully submitted (valid server response) via ajax or a regular POST request. While it might be nice to have an enhancement that automatically adds the success class to the form after a form is submitted and the server responds with an OK, I will leave that out of this pull request.

fixes #3683


jlukic:
You can modify the className 'success' in the settings object to set as a blank string. Removing this feature would cause code for those that already rely on this feature to break, so this could not be changed without a major release, and I'm happy currently with the decision to use it.

PR #4614, opened 2016-10-01
Take content settings before attr title

aaronbhansen:
Since settings starts off as false, if it was passed in, it should be taken before title is returned. Workflow should still stay the same, since settings.content false would still default to title.


jlukic:
This is just plain wrong, thanks.

jlukic:
And a zillion years later, fixed

championswimmer:
yay !!

PR #4631, opened 2016-10-06
Fix inverted dimmer text color

rijk:
The definition references the wrong variable: `@textColor` instead of `@invertedTextColor` which is defined in https://github.com/Semantic-Org/Semantic-UI/blob/ebb1cab098cad4509d8b173bd1d185b7472711f6/src/themes/default/modules/dimmer.variables#L54.

### DIFF
--- a/src/definitions/modules/dimmer.less
+++ b/src/definitions/modules/dimmer.less
@@ -153,7 +153,7 @@ body.dimmable > .dimmer {
   background-color: @invertedBackgroundColor;
 }
 .ui.inverted.dimmer > .content > * {
-  color: @textColor;
+  color: @invertedTextColor;
 }
 
 /*--------------


jlukic:
In `2.3.0`

lubber-de:
Fixed in https://github.com/fomantic/Fomantic-UI 2.6.4

flekschas:
Unless I am mistaken this is still an issue in `v2.4.0`.

Content within a inverted dimmer is not visible due to the following CSS rule:

```css
.ui.inverted.dimmer > .content > * {
    color: #FFFFFF;
}
```

Here's an example: https://jsfiddle.net/7x2hr1py/

Or am I doing something wrong?

lubber-de:
@flekschas Yes, you are right. Semantic UI still has this issue. It is only fixed in the above mentioned Community Fork

flekschas:
@lubber-de Awesome thanks for the quick reply. I just replied in case other people come across the issue and assumed it's fixed because it's listed under `v2.3.0` :)

PR #4651, opened 2016-10-14
Fix #3424

### LINKED ISSUE #3424: [Dropdown][Search Selection] Match any part of the string
This is a feature request.

Currently, when you search in a dropdown it matches the string from the beginning only. You can't search for something in the middle of the string (LIKE% matches, %LIKE% doesn't). For example, if you want to search a dropdown of users and you type their last name but no results are found unless you type the first name first.

#### Example
http://semantic-ui.com/modules/dropdown.html

#### What happens
- Go to the country dropdown search select.
- Type **states**
- No results found.

#### What's expected
- Go to the country dropdown search select.
- Type **states**
- **United States** shown in the list as a match.

#### Example of expected
jQuery's autocomplete does this and there is an example here: https://jqueryui.com/autocomplete/#combobox

rminnett:
Dropdown fullTextSearch=exact was only implemented for the text search, not the value search. The value term was using a fuzzy match even with fullTextSearch=exact.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -789,12 +789,15 @@ $.fn.dropdown = function(parameters) {
                 }
                 if(settings.match == 'both' || settings.match == 'value') {
                   value = String(module.get.choiceValue($choice, text));
-
                   if(value.search(beginsWithRegExp) !== -1) {
                     results.push(this);
                     return true;
                   }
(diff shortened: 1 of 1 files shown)


jlukic:
This is incorrect as stated. Marking to a release.

rminnett:
Great, thanks!

Issue #4506, opened 2016-08-30
[Dropdown] - Search dropdown enter and blur couses first option selection


michtill:
**Steps to Reproduce**
Enter and blur search selection dropdown with yet selected item.
1. Select some "not first" item in the search dropdown.
2. Again click into dropdown (in the search input, not only the arrow).
3. Click anywhere else outside the dropdown (blur the dropdown).

**Expected**
The dropdown should keep previesly selected item.

**Result**
The dropdown selects the first item in the list.

**Testcase**
http://jsfiddle.net/ba7dkjx4/

**WORKAROUND**
forceSelection: false //but on search dropdown it should be default false

clthck:
I can confirm this bug

gizzard:
The jsfiddle testcase is empty. I created one example here:
http://jsfiddle.net/0hfpccjz/

forceSelection: false is not an easy workaround for me because this allows to set a string which is not in the dropdown.

brunotourinho:
Confirmed!

khornberg:
I've gotten this as well

khornberg:
Setting `forceSelection` to `false` solved the problem for me.

juice:
This Issue should be included into the next Release.

MaikeMota:
We got the same issue in here.

jlukic:
Disable `selectOnKeydown` using `selectOnKeydown: false`, this is a separate behavior than whether it forces selection on blur.

juice:
In previous releases it was possible to type a few characters into the dropdown search and confirm the first selected item with tab. What do i have to do to make this work again?

jlukic:
@juice `forceSelection: false`

juice:
@jlukic Hi Jack, thanks for the tip, but this works only 50 percent of the time. I love this project, but right now the Dropdown part is a problem for many week. We use Dropdowns on a page with many Form Fields, most of them are generated from select fields and some of them are remote searches.

People want to type a few characters and tab to the next dropdown or text field. Right now, i have to choose between, dropdowns clear the current selection/add the first item to a multiselect or tab select is only working for some dropdowns if "forceSelection: false" is set, plus the typed characters are rendered over the selected text.

[...]

PR #4656, opened 2016-10-17
#4506 do not change selection when previously selected element is unfiltered

cbxp:
firstUnfiltered() forced change of selection even if previously selected element was visible. This meant that when tabbing through the field, the selected option was changed to first visible.


jlukic:
The bug this referenced previously is closed. Can you provide a jsfiddle for what this is fixing?

jlukic:
I'm still interested in looking at this if you can provide a test case

PR #4678, opened 2016-10-22
Slider: bugfix and improving implements

flytreeleft:
There are following changes:
- Code cleaning and format;
- Add slider container padding when locating labels;
- Use css function `calc()` to locate thumb, track fill and labels, so no need to listen resize event or observe element changes;
- Extra parameter `thumbVal`, `secondThumbVal` for listener `onChange()` and `onMove()`;
- Exactly updating global variables `value`, `module.thumbVal`, `module.secondThumbVal`, `position` and `secondPos`;
- Set or update `value` to `min` or `max` when mouse is moving out slider;

### DIFF
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -342,7 +342,7 @@ $.fn.slider = function(parameters) {
           move: function(event, originalEvent) {
             event.preventDefault();
             var value = module.determine.valueFromEvent(event, originalEvent);
-            if(module.get.step() == 0 || settings.smooth) {
+            if(module.get.step() == 0 || module.is.smooth()) {
               var
                 thumbVal = module.thumbVal,
                 secondThumbVal = module.secondThumbVal,
@@ -456,9 +456,12 @@ $.fn.slider = function(parameters) {
           vertical: function() {
             return $module.hasClass(settings.className.vertical);
           },
+          smooth: function() {
+            return settings.smooth || $module.hasClass(settings.className.smooth);
+          },
           touch: function() {
             return isTouch;
-          },
+          }
         },
 
         get: {
@@ -1182,7 +1185,8 @@ $.fn.slider.settings = {
     disabled : 'disabled',
     labeled  : 'labeled',
     vertical : 'vertical',
-    doubled  : 'double'
+    doubled  : 'double',
+    smooth   : 'smooth'
   },
 
   keys : {
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -48,8 +48,8 @@ $.fn.slider = function(parameters) {
 
       var
         settings        = ( $.isPlainObject(parameters) )
-          ? $.extend(true, {}, $.fn.range.settings, parameters)
-          : $.extend({}, $.fn.range.settings),
+          ? $.extend(true, {}, $.fn.slider.settings, parameters)
+          : $.extend({}, $.fn.slider.settings),
 
         className       = settings.className,
         metadata        = settings.metadata,
--- a/src/definitions/modules/slider.js
+++ b/src/definitions/modules/slider.js
@@ -165,8 +165,9 @@ $.fn.slider = function(parameters) {
             $thumb = $module.find('.thumb:not(.second)');
             $currThumb = $thumb;
             if(module.is.doubled()) {
-              if($module.find('.thumb.second').length == 0)
+              if($module.find('.thumb.second').length == 0) {
                 $module.find('.inner').append("<div class='thumb second'></div>");
+              }
               $secondThumb = $module.find('.thumb.second');
             }
             $track = $module.find('.track');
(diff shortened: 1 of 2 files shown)


gdaunton:
Found a bit of an issue when I set the min to something other than 0
untitled

src/definitions/modules/slider.js
```diff
@@ -193,48 +201,39 @@ $.fn.slider = function(parameters) {
...
                 var
-                  $label = $('<li class="label">' + module.get.label(i+1) + '</li>'),
-                  position =
-                    module.is.vertical()
-                    ?
-                    module.is.reversed() ? 'bottom' : 'top'
-                    :
-                    module.is.reversed() ? 'right' : 'left'
+                  $label = $('<li class="label">' + module.get.label(i) + '</li>'),
+                  ratio  = i / (len + 1)
```
gdaunton: this isn't quite right. If I have a max of 10 and a min of 0, I am going to have 8 labels. So this would be `1/9` not `1/10` as it should be.

With the change to `module.get.numLables()`: `i / (len + 1)` -> `i / len`

src/definitions/modules/slider.js
```diff
@@ -193,48 +201,39 @@ $.fn.slider = function(parameters) {
...
-                $labels.empty()
-              else
+              if($labels.length != 0) {
+                $labels.empty();
+              }
+              else {
                 $labels = $module.append('<ul class="auto labels"></ul>').find('.labels');
-              for(var i = 0; i <= module.get.numLabels(); i++) {
+              }
+              for(var i = 1, len = module.get.numLabels(); i <= len; i++) {
```
gdaunton: `i <= len` -> `i < len` to go along with the change in `module.get.numLables()`:

```javascript
var value = Math.round((module.get.max() - module.get.min()) / module.get.step()) - 2;
```
should now be
```javascript
var value = Math.round((module.get.max() - module.get.min()) / module.get.step());
```

src/definitions/modules/slider.js
```diff
@@ -193,48 +201,39 @@ $.fn.slider = function(parameters) {
...
-              ratio,
-              position
+              min         = module.get.min(),
+              max         = module.get.max(),
+              ratio
             ;
             $children.each(function(index) {
               var
-                $child = $(this),
+                $child    = $(this),
```
gdaunton: nit: no need for more whitespace

flytreeleft:
@gdaunton Thanks, all known issues are already fixed. :)

gdaunton:
Found an issue with the double handle:
i1rjll3hrc

It works fine after I grab the second handle.

gdaunton:
Really love the calc solution on this 👍 . 

I think it is a tiny bit slower, but it does get rid of the positioning issues when the slider width changes.

Thanks for the help!

src/definitions/modules/slider.js
```diff
@@ -122,7 +122,7 @@ $.fn.slider = function(parameters) {
           module.unbind.events();
           module.unbind.slidingEvents();
           $module.removeData(moduleNamespace);
-          module.disconnect.sliderObserver();
+          // module.disconnect.sliderObserver();
```
gdaunton: Go ahead and remove this line and function if it is no longer used

src/definitions/modules/slider.js
```diff
@@ -244,7 +243,7 @@ $.fn.slider = function(parameters) {
         bind: {
           events: function() {
             module.bind.globalKeyboardEvents();
-            module.bind.resizeListener();
+            // module.bind.resizeListener();
```
gdaunton: 🔪

src/definitions/modules/slider.js
```diff
@@ -309,7 +308,7 @@ $.fn.slider = function(parameters) {
             $module.off('touchstart' + eventNamespace);
             $module.off('keydown' + eventNamespace);
             $module.off('focusout' + eventNamespace);
-            $(window).off('resize' + eventNamespace);
+            // $(window).off('resize' + eventNamespace);
```
gdaunton: 🔪

src/definitions/modules/slider.js
gdaunton: Also you can probably remove the resize listener

src/definitions/modules/slider.js
```diff
@@ -103,9 +103,9 @@ $.fn.slider = function(parameters) {
           module.read.metadata();
           module.read.settings();
 
-          module.observeChanges();
+          // module.observeChanges();
```
gdaunton: 🔪

src/definitions/modules/slider.js
```diff
@@ -138,6 +138,9 @@ $.fn.slider = function(parameters) {
           slider: function() {
             sliderObserver.observe($module[0], {
```
gdaunton: Do we still need to have an observer? I can't think of anything in the class or style that would change where we would need to re-sync everything as all the positioning is percentage based.

src/definitions/modules/slider.less
```diff
@@ -200,14 +200,15 @@
...
+  height: auto;
   left: 50%;
   top: 0;
+  bottom: 0;
   transform: translateX(-50%);
 }
 
 .ui.labeled.vertical.slider > .labels .label {
-  transform: translate(-100%, -@labelWidth);
+  transform: translate(-100%, -50%); // translate(-100%, -@labelWidth);
```
gdaunton: you can 🔪  the comment

flytreeleft:
@gdaunton The commented codes are all removed.

And, I create a [demo](https://jsfiddle.net/flytreeleft/u0nfconw/3/) which can be used to test new changes. The issue you mentioned should already be fixed in previous submits :)

PR #4705, opened 2016-11-03
[Button] Make text visible in loading button and add without text option

### LINKED ISSUE #2271: loading button text suggestion
It'd be nice to be able to have some text next to a loading indicator inside a ```.loading.button```
http://beta.semantic-ui.com/elements/button.html#loading

eppfel:
This fixes #2271, but breaks backwards compatibility.

So, maybe this should be implement the other way around, with ".with .text", but I felt overwriting `color` with the `!important `is already bad practice.


eppfel:
@jlukic  Can you have a look? I might have to revise it...

PR #4707, opened 2016-11-04
Fix <body> being set to position relative when dimmable class applied

jinyangzhen:
Found body is being set to position:relative and cause some of our JS code function incorrectly for example drag and drop. The root cause seems to be this style. I assume there is a typo here.


jlukic:
That appears to be a typo :/

jlukic:
Thank you for your patience here

jlukic:
Fix added in `2.2.11`

Issue #3801, opened 2016-03-08
[Segment] Add basic for segments


esargent:
Enhancement or Bugfix depending on how you look at it:

When you use a `basic segment` (because you don't want a border) inside a `horizontal segments` class, you still get the border because the `.ui.horizontal.segments > .segment` rule takes precedence. So I recommend adding a `basic.segments` definition to solve the problem and maintain the same expectation even if a `basic segment` is wrapped by a `segments` element of some type.

```LESS
/*-------------------
   Basic
--------------------*/`

.ui.basic.segments {
  background: @basicBackground;
  box-shadow: @basicBoxShadow;
  border: @basicBorder;
  border-radius: @basicBorderRadius;
}
```

usage example:

```HTML
<div style="background-color:black;"> <!-- because you can see the issue clearly -->
    <div class="ui basic horizontal segments"> 
        <!--  below basic will be overridden by .ui.horizontal.segments > .segment -->
        <div class="ui basic segment">some content</div> 
        <div class="ui basic compact segment">Free radicals are bad.</div>
    </div>
</div>
```

Adding as an issue vs pull  because I'm not a CSS expert and not clear this is the best way to fix or if it might have other side effects.

kmd1970:
I second this! Had to create my own custom css.

```css
/* remove shadows from segment ouline, basic segment? */
.ui.no.shadow.segment {
 	box-shadow: none;
}
```

jokeronaldo:
Thanks for the sugestion. We will consider this enhancement.

sambro:
I looked through the "New in 2.2" page and didn't see any mention of this enhancement. Did I miss it? Or is this not going to be an enhancement after all? Thank you. :bowing_man:

Banandrew:
Not sure why it was closed, I’ll reopen it, because there’s currently one duplicate of this issue that I’d like to reroute here. If that’s a mistake, please close both of them as rejected.

lubber-de:
Available in https://fomantic-ui.com/elements/segment.html#basic

lubber-de:
Available in https://fomantic-ui.com/elements/segment.html#basic

PR #4773, opened 2016-11-22
[Segment] A basic segment within .segments should be basic

shadoWalker89:
This PR try to fix a problem and add a new functionality :

- The fix : when we have a `.ui.basic.segment` inside of `.ui.horizontal.segments`, the basic segment doesn't loose the border and shadow.

- The new functionality : Add a `.ui.basic.segments` with **s**
It comes in handy when we want to do something like this
```css
.ui.horizontal.basic.segments>(.ui.basic.segment+.ui.basic.right.aligned.segment)
```

### DIFF
--- a/src/definitions/elements/segment.less
+++ b/src/definitions/elements/segment.less
@@ -463,7 +463,9 @@
        Basic
 --------------------*/
 
-.ui.basic.segment {
+.ui.basic.segment,
+.ui.segments .ui.basic.segment,
+.ui.basic.segments {
   background: @basicBackground;
   box-shadow: @basicBoxShadow;
   border: @basicBorder;


shadoWalker89:
@jlukic bump

hammy2899:
@shadoWalker89 Can you make a jsfiddle to replicate the issue?

shadoWalker89:
@hammy2899 I updated the issue with the jsfiddle link
The expected behaviour is that the segments do not have borders because they are using .basic

dlynch29:
I have this issue too. I can't create basic horizontal segments.

lubber-de:
Fixed. See https://github.com/Semantic-Org/Semantic-UI/issues/3801

JonSilver:
Genuinely confused. This functionality isn't anywhere to be seen in 2.4.2. Are you implying it's been added?

lubber-de:
@JonSilver
It is fixed / implemented in Fomantic-UI as commented in the above mentioned issue
https://fomantic-ui.com/elements/segment.html#basic

Issue #4780, opened 2016-11-22
[Header] Icon in the header: wrong vertical alignment


nZeus:
Here is a fiddle where you can see the problem:
https://jsfiddle.net/Kpblca4/8z2a0fgm/

When using a header with an icon inside, icon is not properly aligned. For the end-user it seems that the icon is lower than the text. I've found that [this line](https://github.com/Semantic-Org/Semantic-UI/blob/f725b162e70896e38257965424ac7f9af486b927/src/definitions/elements/header.less#L67) can be removed.

sbmelvin:
I've encountered the same alignment problem when using icons in list items #4872.

sbmelvin:
Actually, I just noticed that this is not the same (although related). My issue was was the horizontal alignment, not vertical. My fix was setting padding-right: 0; Perhaps the pull request by @ivannieto could be amended to include padding-right:0 as well?

philrykoff:
In #4872, there is a solution to this issue.

PR #4850, opened 2016-12-22
Fix for #4780

ivannieto:
Added padding-top: 0px to avoid wrong placement of the icon as stated by the fix in the fiddle from the issue comment.


ivannieto:
Seems to be fixed in #4872. Closing.

jlukic:
Fixed in next release. Removed the `@lineHeightOffset`

PR #4910, opened 2017-01-14
[Slider] Fixes for docs

gdaunton:
A few fixes for the docs so that they run correctly.

Also added a few things:
- Custom label interpreter function
- Bug with reversed slider label positioning

/cc @jlukic @flytreeleft


flytreeleft:
@gdaunton The variable `sliderObserver`(which is at line 84 in `src/definitions/modules/slider.js`) should be removed.

I didn't ever notice it in my last PR :(

Issue #4479, opened 2016-08-25
[Modal] Prevent modal actions buttons from clicking twice


kbariotis:
Modal contains transitions to showing on the screen. Action buttons are able to be clicked more than once while a modal is transitioning.

Here's an example http://jsfiddle.net/rajsnzxw/10/. Open console and try to click the button more than once aggressively.

I was wondering whether this behavior is something that Semantic's code must handle. In my current project, there is no way to fix this issue globally without adding side effects on my code base.

Thank you

jlukic:
Yes, we should probably handle rapid clicking on a framework level.

jlukic:
Fixed in `next`

kmd1970:
Unfortunately this is **not** fixed in 2.2.10

https://jsfiddle.net/kmd1970/8joakfjh/

lubber-de:
@kmd1970 It is actually fixed. But your fiddle is slightly wrong: You've bound the show trigger to **all** buttons including the buttons from the modal itself: See your fixed fiddle here: https://jsfiddle.net/16o8Ly0n/

PR #4933, opened 2017-01-22
Fix for [Modal] Prevent modal actions buttons from clicking twice #4479

ob6160:
Problem is described here: #4479

Adds a check for whether the modal is animating before firing the approve event. (Line 212)

Additionally adds a check for the `show` function in order to ensure that the modal is not animating and is not active before allowing it to be shown again. (Line 311)


jlukic:
I've added a slightly different solve that only specifically changes behavior when `onApprove` or `onDeny` callbacks are used. The changes to show would cause issues for anyone who expects `show` to reshow a modal while the current one is hiding.

	91dfe87

PR #4957, opened 2017-01-29
Use '===' comparison

dijonkitchen:
Found via [HoundCI](https://houndci.com/)'s error I get after a `gulp build`.

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1523,30 +1523,30 @@ $.fn.form.settings = {
     },
 
     minCount: function(value, minCount) {
-      if(minCount == 0) {
+      if(minCount === 0) {
         return true;
       }
-      if(minCount == 1) {
+      if(minCount === 1) {
         return (value !== '');
       }
       return (value.split(',').length >= minCount);
     },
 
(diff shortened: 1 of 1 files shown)


dijonkitchen:
I think this is still relevant!

dijonkitchen:
Can this be merged? It's still relevant.

brodybits:
??

dijonkitchen:
@jlukic ?

jlukic:
@dijonkitchen 

This PR breaks the functionality of the rules you've modified. 

Values are specified like `minLength[1]` and then split into `minLength` and `1` via regex, see `get.ancillaryValue` below for how this is parsed.

```javascript
ancillaryValue: function(rule) {
  if(!rule.type || (!rule.value && !module.is.bracketedRule(rule))) {
    return false;
  }
  return (rule.value !== undefined)
    ? rule.value
    : rule.type.match(settings.regExp.bracket)[1] + ''
  ;
},
```

Using `===` comparison would cause the rules conditions to never be run. 

---------------------

@dreaming-augustin Please be mindful of your tone. It's not appropriate to tag onto PRs to share negative feedback about the project.

Issue #4990, opened 2017-02-06
[Dropdown] Add 'filter' setting to allow 'contains' vs 'startsWith' regex


damonsutherland:
In the semantic.js line 4970 (in the function filterItems) sets a RegExp to '^' + escapedTerm.  It would be nice to have, at a minimum, the options of 'startsWith' or 'contains'.  It could be as simple as:
`
beginsWithRegExp = settings.filter === 'startsWith' ? new RegExp('^' + escapedTerm, 'igm') : new RegExp(escapedTerm, 'igm');
`

or we could add another value to fullTextSearch, i.e. fullTextSearch === 'contains' and modify accordingly.

PR #4992, opened 2017-02-06
4990: Adding filter setting for dropdown; default is 'contains'

damonsutherland:


Banandrew:
Linking the initial issue here: #4990.

PR #5022, opened 2017-02-15
[Grid] Add fitted column to equal width grids

hugopeek:
This small change allows you to add the class name "fitted" to any column inside an equal width grid. A fitted column will wrap around its contents by setting flex-grow to 0.

Here's [a fiddle](http://jsfiddle.net/k7u4myq3/4/) to demonstrate.

A small override (width: auto) is needed to prevent a default width from being applied to the column.


Giwayume:
`fitted` is used to mean "lack of padding" elsewhere in the framework. I suggest calling it `compact` to keep in line with naming conventions.

hugopeek:
@Giwayume Good idea, I changed it to 'compact'.

Issue #5053, opened 2017-02-21
[Modal] Make modal ARIA compliant with role="dialog" and fix keyboard accessibility issues.


Giwayume:
One huge thing this framework is missing is accessibility compliance. 

A few issues:
* role of dialog is missing
* aria-labelledby missing (assuming this could always be the header)
* first focusable control should be focused when modal is shown
* keyboard focus should be resumed on the last focused element before the modal was shown, after the modal is hidden
* tabbing order should wrap inside the modal and not allow focus to escape the modal while open

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role

Giwayume:
The above was based on the current release. I see after downloading the following items have already been addressed:
* first focusable control should be focused when modal is shown
* keyboard focus should be resumed on the last focused element before the modal was shown, after the modal is hidden

PR #5058, opened 2017-02-22
[Modal] #5053 Focus trapping inside the modal 

Giwayume:
This code is ready for review.

From issue #5053, implements focus trapping inside of the modal. Tabbing from the last focusable element will move to the first inside the modal, and vice a versa.

Complies with ARIA accessibility guidelines.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role


Giwayume:
I meant to do a cross repository merge; disregard.

PR #5095, opened 2017-02-28
[Checkbox] Added `box-shadow` to the checkbox component’s `.toggle` variation

### LINKED ISSUE #5094: Toggle checkbox no longer has box-shadow
Starting at 2.2.5 the Checkbox "toggle" style has no longer included a box-shadow by default, which (according to our users) makes it hard to see and when seen appear disabled. It's also now inconsistent with the "slider" checkbox, which still has the drop shadow.

#### 2.2.4
image

#### 2.2.5
image

https://jsbin.com/wiretu/edit?html,output

I'm guessing this might have been accidental, instead intending to just allow it to be configurable? I tried to find the true, original PR but seems it was part of some sort of complicated fork merge.

https://github.com/Semantic-Org/Semantic-UI/compare/2.2.4...2.2.5#diff-46cda3168af13ade8aea3ab4a1edd16cR522

You can get the old style back by using:

```less
/* your-semantic/site/modules/checkbox.variables */
@toggleOffHandleBoxShadow: @handleBoxShadow;
```

Banandrew:
Resolves #5094.

### DIFF
--- a/src/definitions/modules/dropdown.less
+++ b/src/definitions/modules/dropdown.less
@@ -182,7 +182,7 @@
 }
 
 .ui.dropdown .menu > .input {
-  width: auto;
+  width: auto !important;
   display: flex;
   margin: @menuInputMargin;
   min-width: @menuInputMinWidth;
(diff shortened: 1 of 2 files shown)


Banandrew:
Forgot to branch out.

PR #5122, opened 2017-03-03
More flexible creditCard validation

neokio:
You've done a spectacular job of credit card validation.
But the code forces unnecessary conditions: 
1) Users must enter credit cards without dashes
2) UX improvements which allow user-friendly credit card entry are by design invalid (such as [js formatter](https://github.com/mypebble/formatter.js) ... demo [here](http://firstopinion.github.io/formatter.js/demos.html))

I've added one line of code to strip non-numerics before validation, returning control to the developer.

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -1520,6 +1520,8 @@ $.fn.form.settings = {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


jlukic:
I do agree dashes are reasonable to include in credit card strings. I dont believe we should validate other strings as correct though (a-Z etc).

I'll add with a slight change.

Issue #3214, opened 2015-10-20
[Modal] Add tiny and mini modal variations.


ismail-codar:
Large and small modal sizes can be applied #240 but noting for mini, tinny and other sizes.
Small modal  size is 720px on big screens. It is big for name, surname like inputs in the modal.

Banandrew:
In the meantime, it doesn't take too much time to create your own; this example should look nicer with forms in general.

*src/site/modules/modal.ovverides:*

```css
/* Extra Small */
.ui.xsmall.modal > .header:not(.ui) {
  font-size: @xsmallHeaderSize;
}

/* Extra Small Modal Width */
@media only screen and (max-width : @largestMobileScreen) {
  .ui.xsmall.modal {
    width: @xsmallMobileWidth;
    margin: @xsmallMobileMargin;
  }
}
@media only screen and (min-width : @tabletBreakpoint) {
  .ui.xsmall.modal {
    width: @xsmallTabletWidth;
    margin: @xsmallTabletMargin;
  }
}
@media only screen and (min-width : @computerBreakpoint) {
  .ui.xsmall.modal {
    width: @xsmallComputerWidth;
    margin: @xsmallComputerMargin;
  }
}
@media only screen and (min-width : @largeMonitorBreakpoint) {
  .ui.xsmall.modal {
    width: @xsmallLargeMonitorWidth;
    margin: @xsmallLargeMonitorMargin;
  }
}
@media only screen and (min-width : @widescreenMonitorBreakpoint) {
  .ui.xsmall.modal {
    width: @xsmallWidescreenMonitorWidth;
    margin: @xsmallWidescreenMonitorMargin;
  }
}
```

*src/site/modules/modal.variables:*

```LESS
/*******************************
         Site Overrides
*******************************/

/*-----------------------------
      Modal Variations
-----------------------------*/

/* Size Widths */
@xsmallRatio: 0.4;

/* Derived Responsive Sizes */
@xsmallHeaderSize: 1.3em;
@xsmallMobileWidth: @mobileWidth;
@xsmallTabletWidth: (@tabletWidth * @xsmallRatio);
@xsmallComputerWidth: (@computerWidth * @xsmallRatio);
@xsmallLargeMonitorWidth: (@largeMonitorWidth * @xsmallRatio);
@xsmallWidescreenMonitorWidth: (@widescreenMonitorWidth * @xsmallRatio);

[...]

PR #5123, opened 2017-03-03
Added a `.tiny` modal size

Banandrew:
It seems to be a common request: #3214. Feel free to modify the variables; I’ve been using these settings for quite a while, and considering the optimal line length of the default theme’s font size, `0.4` ratio works great with small forms, messages or text. Not so sure about `@tinyHeaderSize` that I kept at `1.3em`, though.

I’ll submit a PR to the documentation as soon as I figure out how to add JavaScript to it.

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -408,6 +408,43 @@
   font-size: @medium;
 }
 
+/* Tiny */
+.ui.tiny.modal > .header:not(.ui) {
+  font-size: @tinyHeaderSize;
+}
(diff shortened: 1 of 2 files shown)


Banandrew:
I couldn’t run the documentation locally, and while it’s very easy to update it to add the `.tiny` variation example, I’d rather not open an unproven PR.

jlukic:
I'll update.

jlukic:
Added `tiny` and `mini` in `2.1.11`

PR #5148, opened 2017-03-10
Fixed an issue with child `.card` elements inheriting their parents’ rules

### LINKED ISSUE #5128: [Icon] `.id.card.icon` and `.address.card.icon` inherit the card component’s rules
I think this is because the "card" in the icon name tricked the "cards" wrapper to believe it is a card object.

JSFiddle: https://jsfiddle.net/oL8792km/

Banandrew:
Resolves #5128.


jlukic:
Yep, this is clearly wrong.

PR #5210, opened 2017-03-29
Dropdown text’s error state shouldn’t be applied to menu items that have descriptions

### LINKED ISSUE #5207: [Dropdown] Error style is incorrect when in Form > Field
When using Dropdown inside Form > Field with error class there is a incorrect styling

Referred from semantic ui react: https://github.com/Semantic-Org/Semantic-UI-React/issues/1462
image

Banandrew:
Resolves #5207.

Before:
sui_before

After:
sui_after


lubber-de:
@awgv We adopted your fix in the community fork [Fomantic-UI](https://fomantic-ui.com) by
https://github.com/fomantic/Fomantic-UI/pull/1803

PR #5242, opened 2017-04-06
Fixed dropdown menu’s border having incorrect color under the error state

Banandrew:
Before, it was showing red border only on hover, and I thought it wasn’t correct:
before
After, it shows red border at all times:
after

~Also, I updated my fork to `2.2.10`, and I think I might have messed up something, but although it’s showing 3 commits to the branch, it seems that only one file was changed for `next`—please double check it for me in case I’m missing something.~

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -526,16 +526,9 @@
   color: @formErrorColor;
 }
 .ui.form .fields.error .field .ui.dropdown,
-.ui.form .field.error .ui.dropdown {
+.ui.form .field.error .ui.dropdown,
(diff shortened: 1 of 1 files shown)


sstok:
```
git fetch upstream
git checkout next
git reset --hard upstream/next
git checkout dropdown-error-border
git reset --hard upstream/next
git cherry-pick 5436dae820b3e07ec16834cd455347931ecba195
```

This should do the trick 👍  always prefer to rebase when puling instead of merging as creates ugly merge bubbles.

Banandrew:
Hi @sstok, thank you, that did it. 👍 Will be doing it properly from now on.

PR #5262, opened 2017-04-10
[Grid] Mobile Reversed Stackable Rows should be reversed on mobile

xeophin:
Without the additional selectors, rows could not be set to be mobile reversed and stackable independently from their containing grid.


peterrobertz:
mobile reversed dont work anymore, can someone confirm?

PR #5282, opened 2017-04-17
menu/search: remove border-radius

### LINKED ISSUE #5281: [Menu] Search should not cut off input text
Here's a fiddle demonstrating the issue. I noticed this first in the semantic docs when there's a search input in the menu component so I assume that it's pretty easy to reproduce. The fiddle is basically a copy of the code in the docs.

- Fiddle: https://jsfiddle.net/rrc7an6x/

Here's a screenshot of what I'm getting in case the issue is not re-producable:

image

This seems to be because a `border-radius` on the input, since only the first letter has the cutoff:

image

#### Info
- Browser: Google Chrome Version 57.0.2987.133 (64-bit)
- OS: MacOS Sierra 10.12.4

koddsson:
The `border-radius` property in the search input was cutting of parts of
the first letter in the prompt.

The prompt class name seems to be needed from the example in the docs[1]
so resetting the `border-radius` for a search prompt when it's in a menu
component seems like the most straight-forward solution.

[1]: https://semantic-ui.com/collections/menu.html#search

Closes #5281

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -346,6 +346,10 @@
   width: @verticalImageWidth;
 }
 
+.ui.menu > .ui.search > .ui.input > .prompt {
+  border-radius: 0;
+}
+
 /*******************************
           Coupling
 *******************************/


koddsson:
I wasn't able to get the semantic docs up and running on my machine so this is somewhat untested but just point me to the correct place and I'll add tests and whatnot more needed to get this merged.

[This fiddle](https://jsfiddle.net/koddsson/f2coayv3/) seems to indicate that the fix is good.

PR #5284, opened 2017-04-17
[Input] Fix doubly applied disabled opacity

### LINKED ISSUE #5283: [Input] - Double opacity applied to disabled inputs
**Steps to Reproduce**
1. Create a `.ui.disabled.input` with an `input[disabled]` child.

**Expected**
The `input` should have an opacity rule applied to appear disabled.

**Result**
The `input` has two opacity rules applied and is hardly visible.  This is because the `.ui.disabled.input` wrapper receives `opacity: 0.4` while the `.ui.input input[disabled]` child also receives `opacity: 0.4`.  The result is an input element with an effective opacity of 0.16.

**Testcase**

http://jsfiddle.net/ebecpcLp/2/

image

**Fix**

The `input` element should only have opacity applied if the wrapper is not already also disabled:

**input.less**
```diff
.ui.disabled.input,
-.ui.input input[disabled] {
+.ui.input:not(.disabled) input[disabled] {
  opacity: @disabledOpacity;
}
```

levithomason:
Fixes #5283 

See the issue for details.

### DIFF
--- a/src/definitions/elements/input.less
+++ b/src/definitions/elements/input.less
@@ -82,7 +82,7 @@
 ---------------------*/
 
 .ui.disabled.input,
-.ui.input input[disabled] {
+.ui.input:not(.disabled) input[disabled] {
   opacity: @disabledOpacity;
 }
 


levithomason:
@jlukic Could you add a milestone here?  We're blocked at https://github.com/Semantic-Org/Semantic-UI-React/pull/1591 over this issue.

jlukic:
Added in `next` manually with credit. Sorry about wait, appreciate your patience @levithomason.

levithomason:
Thanks much.

PR #5303, opened 2017-04-24
[Table] #5227 Added missing rules for sortable/inverted headers

### LINKED ISSUE #5227: sorted inverted table has unreadable headers
When I use inverted sortable for a table with standard theme, the headers which are not sorted get unreadable black text on dark background.

Banandrew:
Resolves #5227.

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -930,9 +930,6 @@
 }
 
 /* Inverted */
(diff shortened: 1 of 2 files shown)


Banandrew:
I pushed another commit—it seems that adding an `!important` is enough to make it work, and additional rules are unnecessary.

jlukic:
Thanks @Banandrew

PR #5316, opened 2017-04-26
[Popup] Include margin when calculating positions

streamside:
The popup is incorrectly placed when using these setttings:

* Position is "bottom right"
* The target element has a margin (body in jsfiddle example bellow)

The margin of the parent element is not included when calculating the widhts and heights.

http://jsfiddle.net/8jk6ou0g/1/

### DIFF
--- a/src/definitions/modules/popup.js
+++ b/src/definitions/modules/popup.js
@@ -520,21 +520,21 @@ $.fn.popup = function(parameters) {
               // element which is launching popup
               target : {
                 element : $target[0],
-                width   : $target.outerWidth(),
-                height  : $target.outerHeight(),
+                width   : $target.outerWidth(true),
+                height  : $target.outerHeight(true),
                 top     : targetPosition.top,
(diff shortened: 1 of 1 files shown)


Prabhakar-Poudel:
Also.
I have a navigation bar with caret icon on top right position of the page. I am trying to show a popup to show few navigation links on it. Surprisingly if I resize the window and it starts scrolling the popup moves irrespective of the position it was suppose to be and moves away from the button. What I mean is if my page starts to scroll the popup does not remain fixed once scrollbars comes up and continues to move with the window. Which does not look right.
Please check on that if you can. Thanks!

jlukic:
Hi, this issue is particular with having a `margin` on `body`. I have not been able to reproduce this with any other layout.

In most cases I would say, go with the fix, but adding a body to margin is fairly abnormal since in almost every case body includes the background for the page and would use `padding`. 

I'm going to decline the fix, but thanks for taking the time to put the PR together.

Issue #4196, opened 2016-06-30
[Modal] Resize Causes Scrolling to be Removed


batata004:
Hi,

Many of my customers have been complaining about this bug and I could never find a way to easily reproduce it. Today I decided I should do that cause yesterday more than 40 people complained about this bug. 

To reproduce the bug you need to be using Chrome browser in a computer/notebook at Windows. I am using the most recent version of Chrome and WINDOWS 10. You need to have a screen resolution of at least 1024px (width). Now you just need to open this link:

http://quemfazsite.com.br/temp/bug-semantic-ui.html

After you open it click in the button "CLICK HERE TO OPEN THE MODAL". Ok, fine, a modal will show up. Now click the resize window button of your Chrome browser and resize the width of the window to a smaller one. Now try to scroll down and you cant anymore.

Now, even if you resize your browser window to full width/height you still wont be able to scroll down the modal cause the scrollbars disappeared.

To save you some time I discovered this bug only happens:

1) if you have more than one modal in your code and the buggy modal has to come first in the code (in the code above I have 2 modals but I decided to only show the first one);
2) the content of the buggy modal can be anything that has a tall vertical content which overflows the browser height dimension of the window. In the case above I used a form but you can insert 100's of <br /> inside the modal's content and the bug will still be reproduceable.

I really really hope you can fix this bug cause I use this amazing SEMANTIC UI in many websites and the complains about this problem are just getting bigger and bigger.

philrykoff:
Upon resizing, the body looses the `scrolling `class.

batata004:
@philrykoff thank you so much, I will try that and will come back here to report you if it solved!

batata004:
I tried your code with the most recent version of jquery and it didnt work. I did some changes to it in order to keep compatibility with recent version but it does not work :(

`$(window).on("resize",function() { $("body").addClass('scrolling'); });`

Even with setInterval didnt work. Any other solution Sir? And do you know if some of the guys behind semantic UI will take a look at this bug?

jlukic:
This is a bug. I'll take a peek

anantoghosh:
Ok I think I got the bug.
Referring the example given above http://quemfazsite.com.br/temp/bug-semantic-ui.html
The first modal does not fit the screen while the second does.
Both the modals are modifying the 'scrolling' class on `<body>`.
On resize the first modal sets the 'scrolling' class then the second modal removes it.
[...]

PR #5319, opened 2017-04-26
[Modal] #4196 Disable refreshing for non active/animated modals

p2kmgcl:
This commit solves #4196
When two or more modals coexist on the same page, the refresh method
of each modal is executed, even if one of them is being shown.
By checking if the module is active or being animated, you prevent
conflicts between shown and hidden modals, like the one described in
the issue.


jlukic:
Looks good as gold.

PR #5359, opened 2017-05-10
[Dropdown] fixes not observed changes to select values of select inputs and menus

SierraGolf:
When values of a select input are changed without causing also a label change, the change is not detected and thus the dropdown is out of sync with the underlying DOM.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -163,6 +163,7 @@ $.fn.dropdown = function(parameters) {
           select: function() {
             if(module.has.input()) {
               selectObserver.observe($input[0], {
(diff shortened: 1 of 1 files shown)


SierraGolf:
I just realized that this "fix" breaks multi select boxes, also there is another line where the attributes should be set to true, which is [here](https://github.com/SierraGolf/Semantic-UI/blob/db3a342b8b61a650a0c252a7f5a8aefed5b6ef78/src/definitions/modules/dropdown.js#L2774).

I could get my multi select boxes working by just applying the change with `attributes: true` when the dropdown is not a multi select box, but I think this might have other implications.

I wonder what would be the best approach to fix this problem.

Issue #2355, opened 2015-06-03
[Menu] Support inverted colored pointing menu


ikourfaln:
As you can see:

![menu](https://cloud.githubusercontent.com/assets/9744226/7957188/673dedae-09dd-11e5-9cec-e38eef62e944.png)

It must be:

![673dedae-09dd-11e5-9cec-e38eef62e944](https://cloud.githubusercontent.com/assets/9744226/7957287/067c0338-09de-11e5-8b2f-bf5f91225bc8.png)

jlukic:
Inverted menus currently set active colors with rgba() overlays which cant be matched with dropdown arrows. Will have to adjust css variables to use fixed hex colors for each active color.

ikourfaln:
Thanx for info :+1:

killswitch:
Just wondering what's the status on this? I'm having this issue right now.

![](http://i.imgur.com/rVIm2V0.png)

jlukic:
It's a lower priority fix. You can fix for your site by adjusting `.item .active:after { background-color: #exacthex }`

killswitch:
Okay, I'll see if I can get some time to do this.

jhalbsgut:
It looks like this one has slipped through the cracks. I'm facing the same issue and while I figured out the workaround, as @jlukic mentioned above, I hate doing hacks. I want to fix this for good. I'm attempting to do a PR for it (my first fork and PR ever, so bare with me) and I'm a little bit lost with the source code.  I haven't worked at all with LESS yet, and frankly, I've learned so many new things lately that I can't wrap my head around it right now. Any pointers would be welcome, and I'll do my best at attempting to fix it myself.

Also, if anyone does look into this issue themselves, note that the :after needs to be changed for both when you have a whole menu set to red and also if you specify just one menu item to be red.

Banandrew:
@jhalbsgut Would be great if you fixed it for everybody. Not sure what to say to help you right away, but If you’ll have some questions, I might be able to answer them in our [Gitter chatroom](https://gitter.im/Semantic-Org/Semantic-UI)—I don’t want to bother other participants with constant notifications.

PR #5386, opened 2017-05-19
Fix colors of pointers on inverted colored menu items

hufftheweevil:
Added three more selectors to each color:

```
.ui.inverted.color.menu .active.item:after            //for colored menus
.ui.inverted.menu .color.active.item:after            //for colored menu items (individual)
.ui.inverted.menu .color.active.item:hover:after  //for colored menu items (individual) - hover
```

Also needed to add !important to each to override the standard inverted color.

This is my first PR. And my first time using LESS (although I didn't really do anything LESSy). Comments and alternate suggestions welcome!

Ref: #2355

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -1425,8 +1425,11 @@ Floated Menu / Item
 
 /* Red */
 .ui.inverted.menu .red.active.item,
-.ui.inverted.red.menu {
-  background-color: @red;
+.ui.inverted.menu .red.active.item:after,
+.ui.inverted.menu .red.active.item:hover:after,
+.ui.inverted.red.menu,
+.ui.inverted.red.menu .active.item:after {
+  background-color: @red !important;
 }
 .ui.inverted.red.menu .item:before {
   background-color: @invertedColoredDividerBackground;
@@ -1437,8 +1440,11 @@ Floated Menu / Item
 
 /* Orange */
 .ui.inverted.menu .orange.active.item,
-.ui.inverted.orange.menu {
-  background-color: @orange;
+.ui.inverted.menu .orange.active.item:after,
+.ui.inverted.menu .orange.active.item:hover:after,
+.ui.inverted.orange.menu,
+.ui.inverted.orange.menu .active.item:after {
+  background-color: @orange !important;
 }
 .ui.inverted.orange.menu .item:before {
   background-color: @invertedColoredDividerBackground;
@@ -1449,8 +1455,11 @@ Floated Menu / Item
 
 /* Yellow */
 .ui.inverted.menu .yellow.active.item,
-.ui.inverted.yellow.menu {
-  background-color: @yellow;
+.ui.inverted.menu .yellow.active.item:after,
+.ui.inverted.menu .yellow.active.item:hover:after,
+.ui.inverted.yellow.menu,
+.ui.inverted.yellow.menu .active.item:after {
+  background-color: @yellow !important;
 }
 .ui.inverted.yellow.menu .item:before {
   background-color: @invertedColoredDividerBackground;
@@ -1461,8 +1470,11 @@ Floated Menu / Item
 
 /* Olive */
 .ui.inverted.menu .olive.active.item,
-.ui.inverted.olive.menu {
-  background-color: @olive;
+.ui.inverted.menu .olive.active.item:after,
+.ui.inverted.menu .olive.active.item:hover:after,
+.ui.inverted.olive.menu,
+.ui.inverted.olive.menu .active.item:after {
+  background-color: @olive !important;
 }
 .ui.inverted.olive.menu .item:before {
   background-color: @invertedColoredDividerBackground;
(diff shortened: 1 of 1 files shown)


jhalbsgut:
#2355

Banandrew:
Hi @jhalbsgut, sorry for the delay, here’s a couple of things about it:

1. The commit is currently being merged from `jhalbsgut:master` into `Semantic-Org:master`—it’s better to always branch out from `master` when you work on isolated features or fixes, and then open a PR against  `Semantic-Org:next` branch, which is a requirement as per the repository’s [contributing guidelines](https://github.com/Semantic-Org/Semantic-UI#pull-requests). As a result of this requirement, this PR can’t be merged, but that’s fine, you can either leave it for the maintainer to merge the changes manually, or close it and open a new one. Also, pushing more commits to `jhalbsgut:master` will automatically update this PR.

    You’re branching out to isolate changes from the `master` branch. For example, if you open another PR that’ll be fixing something else, it will include all of the commits from the `master` branch that you probably don’t need, but when you open a PR from an isolated branch, it will include only the commits from that branch. You can call the new branch in your forked repository anything you like; issue number is usually a good choice.

2. Most importantly, I don’t think this fix is enough, [here’s why](http://jsfiddle.net/Banandrew/07ghp300/). Jack [mentioned](https://github.com/Semantic-Org/Semantic-UI/issues/2355#issuecomment-124127077) the addition of a variable for an arrow, because grouping the rulesets fixes only a half of it, so I’d like to propose something presumably better and LESSy.

     Inverted colored variations start at line [1422](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/collections/menu.less#L1422) and generally look like this:
```LESS
/* Red */
.ui.inverted.menu .red.active.item,
.ui.inverted.red.menu {
  background-color: @red;
}
.ui.inverted.red.menu .item:before {
  background-color: @invertedColoredDividerBackground;
}
.ui.inverted.red.menu .active.item {
  background-color: @invertedColoredActiveBackground !important;
}
```
The problem with the current implementation is that it paints the whole menu in a color, and just slightly darkens the `.active` item by layering the transparent black on top of it, so we need to add something like that right below it:
```LESS
/* Red */
.ui.inverted.menu .red.active.item,
.ui.inverted.red.menu {
  background-color: @red;
}
.ui.inverted.red.menu .item:before {
  background-color: @invertedColoredDividerBackground;
}
.ui.inverted.red.menu .active.item {
  background-color: @invertedColoredActiveBackground !important;
}
```
```LESS
.ui.inverted.red.menu .active.item:after,
.ui.inverted.red.menu .active.item:hover:after{
  background-color: #c52424 !important;
}
```
Because of that layering, `#c52424` has to be hard-coded, which makes the whole thing unwieldy—a developer will have to waste some time if he wants to modify the code. LESS has a lot of useful features, and its [color operation functions](http://lesscss.org/functions/#color-operations) can really help us out here, so why don’t we use them to darken the initial color programmatically? We should modify the default theme’s [menu.variables](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/collections/menu.variables):

1. `@invertedColoredActiveBackground` can be removed, because it’s not needed.
2. Line [421](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/collections/menu.variables#L421)—starting right where `@invertedColoredActiveBackground` was just removed—seems to be the right place to place our new variable:
```LESS
/* Inverted Colored */
@invertedColoredDividerBackground: @dividerBackground;
@invertedColoredActiveBackgroundRed: darken(@red, 4%);
```
I chose `@invertedColoredActiveBackgroundRed` as a name, because it will be easier to read when there will be more colors with a similar name right below it. `darken(@red, 4%)` takes a color, `@red` in this case, and darkens it a bit. It also produces a color that’s *not the same* as the one that was used before, or rather, that’s currently being used in `master`, but it’s very similar and nobody will ever notice. Now to change `menu.less` in definitions:
```LESS
/* Red */
.ui.inverted.menu .red.active.item,
.ui.inverted.red.menu {
  background-color: @red;
}
.ui.inverted.red.menu .item:before {
  background-color: @invertedColoredDividerBackground;
}
.ui.inverted.red.menu .active.item,
.ui.inverted.red.menu .active.item:after,
.ui.inverted.red.menu .active.item:hover:after {
  background-color: @invertedColoredActiveBackgroundRed !important;
}
```
Since we can reuse our newly created color for an `.active.item` (also why we don’t need the variable that we removed), we can combine all rules into one block of code. The best way to see changes is to compare them side by side, I guess, but here’s a pretty bad markdown `diff` to make it a bit better as you’re reading this:
```DIFF
/* Red */
.ui.inverted.menu .red.active.item,
.ui.inverted.red.menu {
  background-color: @red;
}
.ui.inverted.red.menu .item:before {
  background-color: @invertedColoredDividerBackground;
}
-.ui.inverted.red.menu .active.item {
-  background-color: @invertedColoredActiveBackground !important;
+.ui.inverted.red.menu .active.item,
+.ui.inverted.red.menu .active.item:after,
+.ui.inverted.red.menu .active.item:hover:after {
+  background-color: @invertedColoredActiveBackgroundRed !important;
}
```
All that’s left is to do the same thing for all other colors. If you have some questions, I’ll be happy to answer.

PR #5400, opened 2017-05-23
Update Edge supported versions

limonte:
https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history

Current stable is 15 :)


jlukic:
Agh, thank you :)

Issue #4983, opened 2017-02-05
Missing icons.woff2 in material ui theme


nakhade:
I've build the material ui theme by changing the theme from default to material for all the elements in theme.config.

When I use that in my application, I get following error message:

Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:5555/css/themes/material/assets/fonts/icons.woff2

I looked at the - semantic/dist/themes/material/assets/fonts folder, but I don't see icons.woff2 generated there.

and I can see that in semantic.css, woff2 is being referenced:

/*******************************
             Icon
*******************************/

@font-face {
  font-family: 'Icons';
  src: url("./themes/material/assets/fonts/icons.eot");
  src: url("./themes/material/assets/fonts/icons.eot?#iefix") format('embedded-opentype'), **url("./themes/material/assets/fonts/icons.woff2") format('woff2'),** url("./themes/material/assets/fonts/icons.woff") format('woff'), url("./themes/material/assets/fonts/icons.ttf") format('truetype'), url("./themes/material/assets/fonts/icons.svg#icons") format('svg');
  font-style: normal;
  font-weight: normal;
  font-variant: normal;
  text-decoration: inherit;
  text-transform: none;
}

Is this a bug? or am I missing something in the build procedure? I cannot get icons displayed when using material ui, but if I change to default theme, then icons are displayed.

Banandrew:
@nakhade can you please elaborate on your directory structure, where did you install SUI, and where’s the compiled CSS ended up being?

nakhade:
Hi @Banandrew 

I built it from the source - v2.2.7
![image](https://cloud.githubusercontent.com/assets/1238896/22641181/89bf950e-ec4d-11e6-8547-353e5e373d09.png)

Added to my application:
![image](https://cloud.githubusercontent.com/assets/1238896/22641170/7fde4aee-ec4d-11e6-973a-6af153ff3850.png)

Error in the console:
![image](https://cloud.githubusercontent.com/assets/1238896/22641145/63072d6e-ec4d-11e6-9b85-bd16b9848571.png)

Banandrew:
@nakhade sorry for the delay, will it work if you copy the default theme’s icons, so that they’ll be available under that path?

Banandrew:
Closing the issue until there’s a reply.

elTallarin:
Building with webpack, I have this same issue and also images are not being copied

elTallarin:
[...]

PR #5405, opened 2017-05-24
Add missing .woff2 file to material theme

bcroq:
icons.woff2 is https://github.com/google/material-design-icons/blob/master/iconfont/MaterialIcons-Regular.woff2


Banandrew:
Resolves #4983.

jlukic:
Cheers

PR #5410, opened 2017-05-25
[Table ] #5406  Fixed nested tables inheriting parents’ styling

### LINKED ISSUE #5406: [Table] Nested tables inherit the parent tables styling
See the following Fiddle:

https://jsfiddle.net/nghL6bc0/

In this fiddle, we have a normal `ui striped table`. Inside one cell, we have a `ui very basic table`. However, this is actually a `ui very basic striped table`, even though I didn't ask for that. This means tables are not composable.

Banandrew:
Resolves #5406.

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -596,13 +596,13 @@
 
 /* Table Striping */
 .ui.striped.table > tr:nth-child(2n),
-.ui.striped.table tbody tr:nth-child(2n) {
+.ui.striped.table > tbody > tr:nth-child(2n) {
   background-color: @stripedBackground;
 }
 
 /* Stripes */
 .ui.inverted.striped.table > tr:nth-child(2n),
-.ui.inverted.striped.table tbody tr:nth-child(2n) {
+.ui.inverted.striped.table > tbody > tr:nth-child(2n) {
   background-color: @invertedStripedBackground;
 }
 


Banandrew:
@ocharles You’re right, there’s a lot more, and not just with `.striped`. It does make sense to crush it all at once, I’ll go through and push another commit as soon as I can.

Banandrew:
@ocharles Updated the PR, I think I got everything. If you notice another nesting issue, please let me know—I tested the table combinations, but it’s absolutely possible that I missed something. I intentionally skipped `.inverted` blocks, because I found that nesting is unsupported by them, and I think it’ll be better to make them aesthetically pleasing first, and deal with nesting afterwards, although, I’m not sure if we need go that far to support every possible use case.

---

@jlukic While going through the file I found two very minor issues that I also fixed:
* `.definition.table` had incorrect `box-shadow` value [here](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/collections/table.less#L258), which resulted in a left border where there should be none, you can see it in [this fiddle](http://jsfiddle.net/Banandrew/fyyzaxyp/) in the lower left corner.
* `.basic.table` [was using](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/collections/table.less#L1009) its own `border-bottom` that wasn’t actually visible. I changed it to `border-top` on `td` elements instead of `tr` like it’s being universally used [by other table variations](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/collections/table.less#L118). `@basicTableCellBorder` was also removed from the default theme, because it’s the only place where it was used. I don’t think it should be kept—while it’s a tiny bit darker than `@rowBorder`, it’s basically indistinguishable.

Opening separate pull requests for them felt unnecessary, hopefully, it’s fine.

ocharles:
Looks really good now. Is there any reason we don't want to do the exact selector everywhere though?

For example, we have other things like:

```css
.ui.table thead th {
```

Is there any reason this *shouldn't* just be `.ui.table > thead > th {`?

Banandrew:
@ocharles Not really, I just couldn’t imagine a use case with tables inside `thead` and `tfoot` elements. It won’t hurt, I guess, I’ll push one more to include that.

ocharles:
Ah, ok. To give you some context, https://circuithub.com/projects/ocharles/eurorack is the page I'm rewriting to use Semantic UI. It's a huge table, and if you click on any table row, you can see that it expands into a lot more detail. For example, the first table cell contains "Placements" and "Specifications" that could both be modelled as tables. Certainly that list of "parts" (when you expand a row) is a table within the overall table. I'm happy to extend this pull request, but not sure I have permission to work on this branch.

Banandrew:
@ocharles It’s alright, I’ll finish it off tomorrow. It will probably be a while before this PR is merged anyway, so I think it’s best to include these changes in your project immediately in `src/site/collections/table.overrides`.

ocharles:
Yep, not a problem!

Banandrew:
@ocharles I pushed another commit if you need it.

ocharles:
It looks good, but there are still things like `.ui.table tfoot tr:first-child > th:first-child` which are too general. That's what I meant about scoping literally everything in this `less` file.

Banandrew:
@ocharles Does it matter? If I didn’t miss anything, all table variations use these general rules, so they don’t affect the presentation of nested tables; scoping them will accomplish nothing.

ocharles:
It matters only in the sense that Semantic UI "infects" non-Semantic UI tables that are nested within a Semantic UI table. Though as said, I don't feel strongly about this.

Banandrew:
Will update the branch to include #5458—linking them here.

lubber-de:
We adopted most of the changes in Fomantic-UI here https://github.com/fomantic/Fomantic-UI/pull/604
Thanks @Banandrew

PR #5507, opened 2017-07-05
Fixed #5366 dropdown menu takes the correct offset into account

### LINKED ISSUE #5366: [Dropdown] module.is.onScreen returns wrong result when context is changed
Have a look at this [fiddle](http://jsfiddle.net/44czvft4/2/). The menu opens above the dropdown and not below as expected.

The reason is that module.is.onScreen uses $currentMenu.offset() for it's calculations. Unfortunately the returned coordinates are relative to the document and not relative to the supplied context. 

I think you need to calculate the offset relative to the supplied context, like this:

```javascript
            calculations = {
              context: {
                scrollTop : $context.scrollTop(),
                offset    : $context.offset(),
                height    : $context.outerHeight()
              },
              menu : {
                offset: $currentMenu.offset(),
                height: $currentMenu.outerHeight()
              }
            };
            onScreen = {
              above : (calculations.context.scrollTop) <= calculations.menu.offset.top - calculations.context.offset.top - calculations.menu.height,
              below : (calculations.context.scrollTop + calculations.context.height) >= calculations.menu.offset.top - calculations.context.offset.top + calculations.menu.height
            };
```

dannyBies:
Fixed #5366 The dropdown menu now takes the correct offset into account by using the offset of the provided context

### DIFF
--- a/dist/components/dropdown.js
+++ b/dist/components/dropdown.js
@@ -3067,7 +3067,6 @@ $.fn.dropdown = function(parameters) {
             calculations = {
               context: {
                 scrollTop : $context.scrollTop(),
-                offset    : $context.offset(),
                 height    : $context.outerHeight()
               },
               menu : {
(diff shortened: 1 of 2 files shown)


jlukic:
I've manually merged this fix into `next`.

jlukic:
Fix in description of #5366

PR #5512, opened 2017-07-06
[Icon] #5468 Add an alias to 500px’s class name

### LINKED ISSUE #5468: [Icon] Change `\35 00px` class name
I'm using semantic ui with Pug js template. And the prolem i got is the class name 500px. They only accept class names begin with "-", "_" or a letter. That really easy to fix, but if a official fix is better.

Banandrew:
Resolves #5468. The commentary below is outdated, and the new class name was added as an alias instead.

Apparently, classes starting with a digit can’t be used in `Pug.js`, so I’d like to suggest changing it from escaped `\35 00px` to `.five.hundred.px` (as [Wikipedia says the brand is pronounced](https://www.wikiwand.com/en/500px)). I understand that it’s a breaking change, but it’s also possible that there’re other libraries that might have an issue with such class names, and maybe it’s better to support these use cases out of the box.

### DIFF
--- a/src/themes/default/elements/icon.overrides
+++ b/src/themes/default/elements/icon.overrides
@@ -712,7 +712,7 @@ i.icon.firefox:before { content: "\f269"; }
 i.icon.opera:before { content: "\f26a"; }
 i.icon.internet.explorer:before { content: "\f26b"; }
 i.icon.contao:before { content: "\f26d"; }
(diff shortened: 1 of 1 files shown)


j-f1:
Why not support both `500px` and `five hundred px`?

Banandrew:
@j-f1 You’re right, thanks; I didn’t think about it for some reason. I pushed another commit to add it as an alias instead.

PR #5522, opened 2017-07-08
[Docs] #5521 Update the JSFiddle to use HTTPS

### LINKED ISSUE #5521: Update the test case JSFiddle to use https resources.
Please set it to use https for the resource by default, it doesn't work with https everywhere.

Banandrew:
Resolves #5521. I also updated jQuery and attached it to my account to be able to change the JSFiddle without the need to update the markdown files. Ideally, I think, SUI should have its own JSFiddle account.


jlukic:
Can you remove the banandrew part of link, should work with just jsfiddle.net/12312

jlukic:
Actually I'll just edit it live

PR #5555, opened 2017-07-17
Update Uglify options

mbaileys:
The preserveComments option was deprecated and no longer works with the latest Uglify in Semantic-UI 2.11. This prevents JS minification with error "GulpUglifyError: unable to minify JavaScript. Caused by: DefaultsError: `preserveComments` is not a supported option".

Fixed according to latest docs : https://github.com/gruntjs/grunt-contrib-uglify#deprecated-options-from-2x


mbaileys:
Merged on wrong branch.

PR #5578, opened 2017-07-25
Fix modal scrolling for overflow

lulalala:
When modal contains absolute positioned element which is taller than available screen, the modal scrolling is not triggered. By using `scrollHeight` the extended height is taken into account.

`scrollHeight` is how Bootstrap modal determines whether to trigger scrolling.

The bug demo:
https://jsfiddle.net/lulalala/ngjvzwkc/1/

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -533,7 +533,7 @@ $.fn.modal = function(parameters) {
 
         cacheSizes: function() {
           var
-            modalHeight = $module.outerHeight()
+            modalHeight = $module[0].scrollHeight;
           ;
           if(module.cache === undefined || modalHeight !== 0) {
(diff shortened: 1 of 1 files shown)


jlukic:
The fix above doesnt work for a few reasons

* When sizes are calculated the modal has `display: none;` 
* Height calculation is used for vertical centering, and offscreen calculation is  based on relative position of modal centering 
* Modifying modal height to use scrollHeight would change the centering 

7450aa3c9c32d0d608e73c2f52b7bfa4f32a718f adds a new loading state for calculating height, and uses both heights together. If `scrollHeight` is larger than modal base height, it'll use scrollHeight to calculate whether the modal is off stage.

lulalala:
thanks for the fix!

PR #5603, opened 2017-07-28
[List] #5585 Increase list item’s specificity to fix inheritance issues

### LINKED ISSUE #5585: [Popup] Inverted popup menu - items on hover turns transparent
Hello,

I am using popup menu with inverted blue classes. The menu appears fine but the menu items on hover vanish (the color of item is same as background)

Here is the issue script
[http://jsfiddle.net/onlyutkarsh/yc884hrd/5/](http://jsfiddle.net/onlyutkarsh/yc884hrd/5/)

Banandrew:
Resolves #5585.

### DIFF
--- a/src/definitions/elements/list.less
+++ b/src/definitions/elements/list.less
@@ -421,7 +421,7 @@ ol.ui.list ol li,
   color: @linkListItemColor;
   transition: @linkListTransition;
 }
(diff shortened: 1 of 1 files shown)


jlukic:
I've also added a fix for the box shadow from menu divider appearing, and additional inheritance for active states and inverted.

PR #5701, opened 2017-08-24
[Dropdown] Adding beforeLabelCreate() hook

quasipickle:
I have a need to be able to access the actual element that is chosen in a multiselect dropdown.  I have data stored in `data-*` attributes that I can't access in `onCreateLabel()`.  This hook is called before `onCreateLabel()`, and is called in the context of the element clicked.


quasipickle:
PR'd into wrong branch.

PR #5739, opened 2017-09-04
[Icon] #5710 Added a `.setting` alias for better compatibility with FA

### LINKED ISSUE #5710: Bug: semantic ui not displaying icon
Hi,

According to documentation semantic ui made a complete port from font awesome 4.7.0. In this version there is an icon named cog, you can see it below:

http://fontawesome.io/icon/cog/

BUT for some reason semantic ui is not displaying this icon. ALL OTHER ICONS from FONT AWESOME that I could test are displaying correctly! Just this icon cog. The icon "cogs" is working fine but the icon "cog" is not. Below is the simplest example showing that this icon is not working. I display 3 icons, 2 are ok an one is not (which is the "cog").

http://quemfazsite.com.br/temp/criar-site123.php

Or I am doing something wrong again? I wouldnt be surprised with so many mistakes I already made here with semantic ui git.

Banandrew:
Resolves #5710.

### DIFF
--- a/src/themes/default/elements/icon.overrides
+++ b/src/themes/default/elements/icon.overrides
@@ -799,6 +799,7 @@ i.icon.cart:before { content: "\f07a"; }
 i.icon.shopping.cart:before { content: "\f07a"; }
 i.icon.bar.graph:before { content: "\f080"; }
 i.icon.key:before { content: "\f084"; }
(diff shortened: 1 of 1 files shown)


Banandrew:
Was fixed with one of the commits that implemented FA 5, so the PR is unnecessary.

PR #5749, opened 2017-09-06
Fix horizontal scrollbar height, closes #5728

### LINKED ISSUE #5728: [Site] horizontal scrollbars are styled oddly, due to style intended only for vertical
Recently some default styling was added for vertical scrollbars, however it doesn't correctly style horizontal ones. The result is horizontal scrollbars are fatter than vertical ones.

image

https://github.com/Semantic-Org/Semantic-UI/blob/796d2f66e26ae5bb76c604fe6a36914ad3d2fe2a/src/definitions/globals/site.less#L135-L138

Because it removes the webkit-appearance, we need to also set the height for horizontal ones. This can be done either by splitting the styles and using the [horizontal/vertical pseudo classes](https://webkit.org/blog/363/styling-scrollbars/) or just by putting a height in the same selector, since (currently) height placed here will only affect horizontal scrollbars, not vertical.

I would have just PR'd this, but I wanted to first confirm what to call the LESS variable? `@customScrollbarHeight`? Is this ambiguous?

jayphelps:
See #5728 for issue details.

***
The [contributing docs say](https://github.com/Semantic-Org/Semantic-UI/blob/master/CONTRIBUTING.md#creating-pull-requests):

> All pull requests should be merged into the `next` branch.

But if I set that as the PR target it appears `next` is 13 commits behind master, so it wants to include those commits too. Is that instruction out of date maybe?

### DIFF
--- a/src/definitions/globals/site.less
+++ b/src/definitions/globals/site.less
@@ -135,6 +135,7 @@ a:hover {
   body ::-webkit-scrollbar {
     -webkit-appearance: none;
     width: @customScrollbarWidth;
+    height: @customScrollbarHeight;
   }
   body ::-webkit-scrollbar-track {
     background: @trackBackground;
--- a/src/themes/default/globals/site.variables
+++ b/src/themes/default/globals/site.variables
@@ -146,6 +146,7 @@
 @useCustomScrollbars: true;
 
 @customScrollbarWidth: 10px;
+@customScrollbarHeight: 10px;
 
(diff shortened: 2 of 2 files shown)


egorshulga:
+1 for the fix
@Banandrew, review it, please

jlukic:
Probably should target this with `:horizontal` pseudo class for clarity

jlukic:
@jayphelps this is in the release going out now, `2.2.14` 

Thanks for your patience. New updates will be quicker in the new year.

PR #5809, opened 2017-09-25
[Dropdown] Fix #5231 Clicking on the dropdown icon will generate multiple hits to the server (when no results are returned)

### LINKED ISSUE #5231: [Dropdown] Clicking on the dropdown icon will generate multiple hits to the server, when no results are returned.
I was just about to upgrade to v2.2.10 when I noticed my remote dropdowns where generating multiple hits to the server when no results are returned. It also looks like caching is ignored.

For some reason this bug only happens if you click on **the actual dropdown icon** (you will see the loading icon spin, stop, spin, stop... ).  

If you look at the console log you will see multiple repeated requests. 

http://jsfiddle.net/kmd1970/pzcm8zo4/ 
(**Please note** this link needs to remain `http://` for the remote ajax call to work properly)

vpeti:
I resolved the loop by adding the noResults message in queryRemote (which is specific to this scenario, so it seemed appropriate) if the length of results is zero

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -758,6 +758,9 @@ $.fn.dropdown = function(parameters) {
                 module.setup.menu({
                   values: response[fields.remoteValues]
                 });
+                if(0 == response[fields.remoteValues].length) {
+                  module.add.message(message.noResults);
+                }
                 callback();
               }
             }


vpeti:
Please accept.

laja:
This is a bugfix, we are also affected. Could someone please look into it?

jlukic:
I've added a solve in next patch, that is similar in style to this PR. Credit in release notes.

lokesh1990:
I am not sure if I should post this as a new bug.
I receive the array as an object. 

So, from this fix,
   `hasRemoteValues = ($.isArray(values) && values.length > 0)`
is set to false because of isArray.

Do you think it could be replaced by 
   `hasRemoteValues = !$.isEmptyObject(values)` ?
Or do you really have to check if it is an array?

My object would look like 
{ 0 : {name: "test", text: "test", value: "101"} }

PR #5818, opened 2017-09-29
[Dropdown] fix #5817 - When an item has the same display value as another item's ID in a multiple search selection dropdown, selecting both will make Semantic UI confused

### LINKED ISSUE #5817: [Dropdown] When an item has the same display value as another item's ID in a multiple search selection dropdown, selecting both will make Semantic UI confused
[See demonstration](https://jsfiddle.net/dt0fuf2d/1/)

1. Add all items
2. Remove the first Angular one

**Observed result**: The last item that shared its display value with the ID was removed, the intended item is no longer removable

**Desired result**: Items can be removed from selection one-by-one as intended

vpeti:


vpeti:
Please accept

laja:
We also need this fixed!

PR #5832, opened 2017-10-04
[Buttons] fix for active basic button text color always having a lower specificity due to !important

jayphelps:
Because the starting style for color includes `!important`, it always wins out against any value you set for `@basicActiveTextColor`.

https://github.com/Semantic-Org/Semantic-UI/blob/796d2f66e26ae5bb76c604fe6a36914ad3d2fe2a/src/definitions/elements/button.less#L705-L708

This PR just adds `!important`, which then makes it win because now the two compete with a level playing field and `.ui.basic.active.button` is higher than `.ui.basic.button`.

***

It's not clear why `!important` is used in semantic, but I imagine it has a legitimate reason and the topic discussed ad nauseam, so I think this is likely the desired solution with the status quo.

image

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -743,7 +743,7 @@
 .ui.basic.active.button {
   background: @basicActiveBackground !important;
   box-shadow: @basicActiveBoxShadow !important;
-  color: @basicActiveTextColor;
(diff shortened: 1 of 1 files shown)


jlukic:
Yep, this is a bug.

Issue #1885, opened 2015-02-27
Colored text


piernik:
How to get colored text ex `class="ui blue text"`

faller:
+1

jlukic:
I understand the desire but I dont plan on adding colored text. I recommend just using css.

JasonRodman:
I think this is one place semantic is lacking behind other frameworks. There needs to be some utility classes that we can use to apply common styles to any element without having to write custom css. Styles that have the same meaning no matter what component you apply time to. Some examples are common text color classes, text alignment, font weight classes, etc. I am sure we could think of a few more if need be. I find myself having to extend semantic to support these since I find them extremely useful since I write less css on a daily basis with them. I would urge you to reconsider your position on this. How do you plan to complete with bootstrap and foundation without at least having feature parity with them? These seem like no brainer bare minimum necessities to me.

ivansglazunov:
+1
http://materializecss.com/color.html
Nice way to work with color
However, there may be too many CSS.
Why do I need it?
Standard CSS colors are very primitive and are not variable.
In design there is always at least 6, and sometimes 10-15 color states that you need to be able to apply to the background and font. The approach did not materialize so flexible.
I think it would be optimal to make a hybrid of:
Bootstrap
(text|bg)-danger (text-red) (text|bg)-warning (text-yellow) ...
Materialize
(text|bg)-(darken|lighten)-1 (12%) 2 (24%) 3 (36%)

destan:
:+1:  colored text is surely a need. Why wouldn't you consider adding it @jlukic ?

Would you merge if somebody make a pr?

hardy2defy:
please add some color

yaboi:
+1

kevspadilla:
+1

chen-ye:
+1

levithomason:
+1

ilikebits:
:+1:

m0nzderr:
+1

Panman8201:
@jlukic I think it's useful so that colors match the theme's / Less colors. Sure, someone could make their own classes, but not having to worry about future color changes is +:100: Please re-consider.

athyuttamre:
+1

stej4n:
+1. It's very annoying. Why is it possible to create a blue button and not a blue text ??

eeng:
+1

Konard:
+1

rpsirois:
+1

[...]

PR #5862, opened 2017-10-15
Add "text-[COLOR]"class to everything in a Less, beautiful, clean way.

johnunclesam:
Copyright @kamalkhan here: https://github.com/Semantic-Org/Semantic-UI/issues/1885#issuecomment-226047499.

I really did not even understand it until today because we can not have the text colors.

This code generates this:

```
.text.white {
  color: #FFFFFF;
}
.text.grey {
  color: #CCCCCC;
}
.text.black {
  color: #1B1C1D;
}
.text.yellow {
  color: #F2C61F;
}
...
```

### DIFF
--- a/src/definitions/globals/site.less
+++ b/src/definitions/globals/site.less
@@ -110,6 +110,14 @@ p:first-child {
 p:last-child {
   margin-bottom: 0em;
 }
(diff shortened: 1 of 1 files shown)


lubber-de:
It's implemented now. See https://jsfiddle.net/63z148gb/

PR #5891, opened 2017-10-25
Category search for a local source

BrockReece:
This will add the ability to use a category search with a local source object

This will find 'bar' from the following structure
And 'foo' if 'name' is passed in with the searchFields array
```js
source: [
  {
    name: 'Foo',
    results: [
      { title: 'bar' },
    ],
  },
],
```


jlukic:
I've implemented this in `next` with a different algorithm. I'm a bit sheepish about adding any recursion in clientside js.

PR #5892, opened 2017-10-25
Added onChanging callback to accordion

GammaGames:
There are callbacks that happen before open and close, so why not change as well?
Well this fixes that!

I added it so that I have a folder icon on the accordion, so I could just add
`onChanging: function() { $('#file_icon').toggleClass("open"); }`
to initialization to make it change the icon to be open or closed before the animation happened, as opposed to after the animation
accordion


jlukic:
Added in `2.3.0`

PR #5893, opened 2017-10-26
Positioning pusher with left and top instead of translate

bunyaminsg:
When `translate` is used in the `pusher`, `position: fixed` is not working for the child elements inside of the `pusher`. Instead of that, using `left: @width` and `top: @height` fixes this problem.


jlukic:
Translate3d is necessary for GPU acceleration, top/left are not gpu accelerated.

PR #5902, opened 2017-10-28
Swap huge and big in menu.variables #5899

### LINKED ISSUE #5899: Menu sizes may have big and huge mixed up
In global variables the sizes go from _big -> huge -> massive_: https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/globals/site.variables#L112

In menu variables they are _huge -> big -> massive_: https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/collections/menu.variables#L446

jeremy091:
Issue #5899 
Closes #5899

Changed menu variables to be _big -> huge -> massive_ to be consistent with global variables.

### DIFF
--- a/src/themes/default/collections/menu.variables
+++ b/src/themes/default/collections/menu.variables
@@ -453,6 +453,6 @@
 @smallWidth: 13rem;
 @mediumWidth: 15rem;
 @largeWidth: 18rem;
-@hugeWidth: 20rem;
-@bigWidth: 22rem;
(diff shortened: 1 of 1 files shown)


jeremy091:
@ktmn 
Thanks for pointing that out. I swapped bigWidth with hugeWidth in the sizes sections.

jlukic:
Thanks for the fix, looks good to me.

PR #5915, opened 2017-11-02
Samsung dropdown mobile. fix #5515.

### LINKED ISSUE #5515: [Dropdown] multiple selection labels not showing up on samsung internet
**Tested on:**

- Android 7 - default samsung internet browser

Using the default samsung internet browser, labels in multiple selection are not showing up.

Troshin2520:
Fixed dropdown issues in Samsung Internet Browser

### DIFF
--- a/dist/components/dropdown.css
+++ b/dist/components/dropdown.css
@@ -317,6 +317,10 @@
   min-width: 100%;
 }
 
+.dropdown a.ui.label {
+  display: inline-block!important;
+}
+
 /*--------------
      Button
 ---------------*/
--- a/dist/components/dropdown.css
+++ b/dist/components/dropdown.css
@@ -318,7 +318,7 @@
 }
 
 .dropdown a.ui.label {
-  display: inline-block!important;
+  display: inline-block;
 }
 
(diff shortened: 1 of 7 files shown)


aguyinmontreal:
fix 5515

m4ikel:
Please check out https://github.com/Semantic-Org/Semantic-UI/pull/5947 which might be a more elegant way of fixing this issue by attaching the clone item to document.body instead. No other changes are necessary in that way. Somehow Samsung Browser is having trouble adding the $module after the select, when you add a alert or timeout all works fine which seems to me the device is actually waiting for the animation to complete, and after it add check if it still exists (which fails).

Troshin2520:
m4ikel, The fix sense in your commit is the same as in mine, but I agree - there's no need in the extra appending. Merged with your changes and committed.

jlukic:
The cloned element has to be the same place in the DOM otherwise other css rules that might affect layout wont be applied. i.e.

`.someOtherClass .ui.dropdown`

aguyinmontreal:
not fixed

PR #5929, opened 2017-11-08
Make stacked step point arrows down

Alamantus:
Also fix step borders by moving it below steps instead of bunching them all up to the right.

This provides a consistent look for the steps when stacked and also remains separate from the "vertical" styled steps.


lubber-de:
We adopted the enhancement for Fomantic-UI here https://github.com/fomantic/Fomantic-UI/pull/

PR #5944, opened 2017-11-11
[Flag] Add England Flag and Alias

zyzniewski:


jlukic:
Wow there was an england flag hiding in there all along.

Issue #5176, opened 2017-03-17
Menu Collection Issue on Samsung Built-in Web Browser, Internet


andygmorris:
It seems that the Menu, including Menu with Dropdown, fails to open on Samsung mobiles' built-in Web Browser (app named Internet). Works fine on Chrome, Firefox, Opera, Safari (mac), etc across multiple devices and OS's, except the Samsung web browser (ie. on Galaxy S7, Note 4, etc)

Example with error seen on **Semantic-UI's** own website :-
[https://semantic-ui.com/collections/menu.html](https://semantic-ui.com/collections/menu.html)
The first three Menu examples do not work (and same for my own-built web applications):
- Browse Courses
- More 
- Spanner
Scroll further down the page, on a Samsung 'Internet' web browser, and you'll find other Menu items where you expect a menu to pop out on a click, just doesn't happen.

Does anyone have any idea how to fix this, explain it, or have a work-around?
Thank you in advance.

andygmorris:
For the record, the Galaxy S7 used to identify the issue was recently updated with Samsung's Nougat 7.0 update. I don't have access to an S7 with Marshmallow (6.x) any more to see if that's related to the issue.

andygmorris:
Still an issue as per Semantic-UI v2.2.10 05-APR-2017.

m4ikel:
This issue is very much critical, going to dive into it today.

m4ikel:
Debug: No animation defined in CSS: scale in

m4ikel:
`$clone = $('<' + tagName + ' />').addClass( elementClass ).appendTo(document.body);`

andygmorris:
@m4ikel Thank you for your efforts. Confirmed as working on the Semantic-UI website now when using a Samsung S7 Internet Browser application. Who now should close this issue as Closed/Resolved??

Banandrew:
Hello everyone.

@m4ikel, could you please open a PR against the `next` branch in this repository?

Once the PR is opened, the issue will stay open until the PR’s merged. Unfortunately, I can’t say when it’s going to happen, because the only one who can do that is the maintainer.

m4ikel:
@Banandrew done https://github.com/Semantic-Org/Semantic-UI/pull/5947

jlukic:
The issue with this fix is that `transition` expects the duplicate element to be in the same position in the DOM so that the CSS rules will apply correctly that determine its final `display` property, like `flex` or `block`. 

[...]

PR #5947, opened 2017-11-12
#5176 Fix bug within Samsung Browser can.transition

m4ikel:
#5176 Fix bug within Samsung Browser can.transition


jlukic:
I left a comment in the related issue, but this solve would cause issues generally with calculating the final display state.

PR #5965, opened 2017-11-20
[Dropdown] Fix IE11 dropdown re-focus issue

AndrPetrov:
@jokeronaldo Here is [JSFiddle](https://jsfiddle.net/ca0rovs3/116/). 

1. Please, use IE11 to reproduce the bug. 
2. Click on user icon and choose `Settings` for example. 

Actual result: dropdown will be closed and immediately opened.
Expected result: dropdown is closing and will not be opened as it is in other browsers.

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -1196,9 +1196,7 @@ $.fn.dropdown = function(parameters) {
                 isBubbledEvent = ($subMenu.find($target).length > 0)
               ;
               // prevents IE11 bug where menu receives focus even though `tabindex=-1`
-              if(module.has.menuSearch()) {
-                $(document.activeElement).blur();
-              }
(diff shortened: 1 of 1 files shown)


jlukic:
I was able to reproduce.

This is caused by using anchor links that point to the same `href`, as a blank link like `href="#"` does. 

I'd recommend not using a blank href link. I can't `preventDefault` here because people can use `href` to link out. I can confirm using an external link wont cause it to re-open before pageload. So this is only a case with anchor tag in page links.

AndrPetrov:
@jlukic JS fiddle is just a sample how to reproduce. I have the same behavior when use regular links in a EmberJS SPA (Single Page Application).

lubber-de:
Fixed since 2.6.4 https://github.com/fomantic/Fomantic-UI/pull/247

PR #5996, opened 2017-12-05
Fix button height bug with in menus

silverwind:
When `.button` is not a direct descendant of `.menu` (for example, when it is wrapped), it would not get this style applied, which caused inconsistent heights of buttons that were wrapped vs. ones that weren't. Removed the `>` operator to fix this.

Refs:

- https://github.com/go-gitea/gitea/pull/3091
- https://try.gitea.io/silverwind/testcopy/commits/branch/master

### DIFF
--- a/src/definitions/collections/menu.less
+++ b/src/definitions/collections/menu.less
@@ -140,7 +140,7 @@
      Button
 ---------------*/
 
-.ui.menu:not(.vertical) .item > .button {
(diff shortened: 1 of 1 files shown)


silverwind:
Turns out this had some side effects, so I'll close it. Recommendation would be to drop the `.ui.menu:not(.vertical) .item > .button` style, but I'm not totally certain that this has no side effects.

PR #6003, opened 2017-12-06
Karma working and node dependency improvments

nmccready:
- karma is no longer expected to be global
- old karma dependencies explicitly set
- jquery monkey patched since the tests are pointing to JQuery 1.X
(size())


karma.conf.js
```diff
@@ -13,9 +13,10 @@ module.exports = function(config) {
     // list of files / patterns to load in the browser
     files: [
       // require jquery
-      'server/files/javascript/library/jquery.js',
+      // 'server/files/javascript/library/jquery.js',
+      './node_modules/jquery/dist/jquery.js',
       // read css from compiled css
-      'docs/build/uncompressed/**/*.css',
+      // 'docs/build/uncompressed/**/*.css',
```
nmccready: This file set does not exist unless gulp produces this.

karma.conf.js
```diff
@@ -72,7 +73,7 @@ module.exports = function(config) {
     // - Safari (only Mac)
     // - PhantomJS
     // - IE (only Windows)
```
nmccready: This can be switched back but Chrome was less of a pain.

package.json
```diff
@@ -14,7 +14,8 @@
     "url": "https://github.com/Semantic-Org/Semantic-UI/issues"
   },
   "scripts": {
-    "install": "gulp install"
+    "install": "gulp install",
+    "test": "karma start --auto-watch=true --single-run=false"
```
nmccready: probably should make this just `karma start`

karma.conf.js
```diff
@@ -72,14 +73,14 @@ module.exports = function(config) {
...
-    browsers: ['PhantomJS'],
+    browsers: ['Chrome'],
 
 
     // If browser does not capture in given timeout [ms], kill it
     captureTimeout: 60000,
 
     // Continuous Integration mode
     // if true, it capture browsers, run tests and exit
```
nmccready: Setting overridden in `package.json` instead.

package.json
```diff
@@ -92,7 +94,15 @@
     "gulp-debug": "*",
     "gulp-git": "*",
     "gulp-tap": "*",
-    "merge-stream": "*"
+    "jasmine-core": "^2.8.0",
```
nmccready: jasmine-core is probably not needed since `karma-jasmine` is so old and bundles it internally.

nmccready:
The entire purpose of this PR is to allow other developers to seamlessly get involved with the unit tests. Developers should not have to guess at what karma dependencies they should install. Therefore, this setup gets karma working as is and it is not global. It should definitely not be global since the karma being used is so old.

Issue #3864, opened 2016-03-23
[Forms] [Feature request] prompt function


xDaizu:
Would it be possible to define rules like this:

```
email: {
                identifier: 'email',
                rules: [
                    {
                        type: 'empty',
                        prompt: options.messages.email_empty
                    },
                    {
                        type: 'email',
                        prompt: options.messages.email_invalid
                    },
                    {
                        type: 'checkRemoteValidity[email]',
                        prompt: foo
                    }
                ]
            },
```

where `foo` is a function called at validation time, which return a string, so when it validates it creates a customized prompt? Picture something like `"The email blabla@example.com is already taken"`. Currently trying that fails with the error `prompt.search is not a function` because of the line:

```
requiresValue = (prompt.search('{value}') !== -1),
```

so maybe in the previous line:

```
 prompt = rule.prompt || settings.prompt[ruleName] || settings.text.unspecifiedRule,
```

we could try calling rule.prompt()? Would that work?

PR #6016, opened 2017-12-11
#3864 Allow a validation prompt to be a function

xDaizu:
Allows the validation prompt to also be a function that returns a string, as discussed in [issue #3864](https://github.com/Semantic-Org/Semantic-UI/issues/3864)


jlukic:
I've implemented this slightly differently in the above commits and have written docs.

PR #6023, opened 2017-12-12
Use configured lookupFields in search 'get result'

willstott101:
Configuring the lookupFields is a core part of the search module configuration. `get result` was not respecting that and had the default hard-coded. As all it does is pass the value through to the `search object` method (where it is an optional argument), I simply removed the reference.

Failing case: https://jsfiddle.net/15xd2e41/1/
Working case: https://jsfiddle.net/ozcwbcf3/1/

### DIFF
--- a/src/definitions/modules/search.js
+++ b/src/definitions/modules/search.js
@@ -424,8 +424,7 @@ $.fn.search = function(parameters) {
           },
           result: function(value, results) {
             var
-              lookupFields = ['title', 'id'],
-              result       = false
+              result = false
             ;
             value = (value !== undefined)
               ? value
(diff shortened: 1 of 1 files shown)


jlukic:
`get result` is also used internally to lookup the correct matching from `inject.id`, this could create false matches on other fields.

willstott101:
Thanks for the review!

I don't see where `get.result` is used in that file at all. It looks to be only a public API method, as far as I can tell.

`inject.id` uses it's own related `inject.result` which seems to be unrelated to me.

I'm not sure what I've missed, but I'd love to get this PR up to scratch if I can.

If it is required internally then perhaps we could add a `searchFields` argument to this method which passes through to `search.object`, and use that argument whenever it's required?

lubber-de:
A PR for Fomantic-UI adopted the provided logic but adds the `id` to the default `settings.searchFields` so 'get result'  behaves the same as before
See https://github.com/fomantic/Fomantic-UI/pull/895

Issue #5713, opened 2017-08-28
[Form Validation] regression: get values with radio returns true or false


kapouer:
I've setup two fiddles, one with semantic-ui 2.2.10 from cdn:
https://jsfiddle.net/xev09j67/2/
and one with semantic-ui 2.2.12 from semantic-ui-css package:
https://jsfiddle.net/mjL46gm4/1/

click on GET and SET buttons to show the problem.

Banandrew:
Hi @kapouer, it seems that it’s working correctly only when you specifically tell it that the fields are connected with `[]` postfix, e.g.: `name="fruit[]"`. Since `[]` postfix only makes sense in the PHP world, and is more of a back-end thing than anything, maybe it’d be better to automatically apply the same mechanism to multiple instances of `name="fruit"` pattern within a single form.

gyuchang:
This is a regression, introduced in 86ea20d5a09463cda6f2e98dacb824c23ee0b592

See [this comment](https://github.com/Semantic-Org/Semantic-UI/commit/86ea20d5a09463cda6f2e98dacb824c23ee0b592#commitcomment-25176800)

kevinresol:
This is pretty critical I would say.

brandonparsons:
Broke my app also.

tincdev:
a quick workaround

in semantic.js: line 1103



                 if(isRadio) {
                    if(values[name] === undefined || values[name] == false ) {
                     values[name] = (isChecked) ? value || true : false;
                  }

kevinresol:
> a quick workaround

This is still not correct. The value should not be a boolean, but a value of the radio element selected.

tincdev:
That what you will have! the value of the selected radio element, but in some case people use radio input without "value" attribute and if you do you will have a boolean value....

kevinresol:
ahhh yes you are right, I mis-read the `||`

jlukic:
I'll look at this over the hoildays.

jlukic:
I've merged in the fix from @tincdev it is the correct approach.

jlukic:
Will be fixed in `2.2.14` released later tonight.

PR #6043, opened 2017-12-19
Fix for [Form Validation] regression: get values with radio returns true or false #5713

tincdev:


jlukic:
I've tested this fix and it appears to be correct. Thanks for your patience.

PR #6047, opened 2017-12-20
Fix #5181 (fluid left/right labeled icon buttons text position)

### LINKED ISSUE #5181: Fluid icon buttons do not render button text in correct position
The text centering for a fluid icon button does not account for the width of the icon.  jsFiddle contrasting behavior of fluid and non-fluid icon buttons:  https://jsfiddle.net/orzu9b78/6/

bcroq:


bcroq:
sorry, I don't know why this pull request is still not merged...

lubber-de:
While testing your fix (and it became part of my testing fiddle) i came across the fact, it's not even neccessary to override the padding, but it's enough to remove the existing padding for the `fluid` variant, because the padding behavior is the same for the non-fluid variant then mathcin your proposed fix,
I did this for Fomantic-UI by https://github.com/fomantic/Fomantic-UI/pull/584

Thanks for pointing me into the right direction 😄

Issue #6067, opened 2018-01-03
gulp build is not building semantic.min.css


nestorvanz:
In a fresh installation i am running "gulp build" but in the "dist" folder is not present semantic.min.css, theres only the semantic.min.js,  "components" and "themes" folder.

wwp66650:
I also have this issue

himanshu230:
I had the same issue
I solved it by changing the build-css gulp task. 
Following are the changes I made:
In file tasks/build/css.js in semantic folder, remove the following line
`compressedStream   = stream.pipe(clone());
 `

nestorvanz:
What are the changes that you made? @himanshu230

himanshu230:
@nestorvanz There is a file <path to your semantic folder>/tasks/build/css.js. Search for line 
`compressedStream = stream.pipe(clone());
`
Removing that line seems to solve the issue. 
It might be because later on, `compressedStream` is again set, which means that this clone stream is not piped to anything, creating backpressure.

jjshammas:
Hmm…that file hasn't changed since July 2016…

Rex90:
Been trying to figure out what I'm doing wrong all morning. Is this an issue with the semantic-ui release?

hammy2899:
@Rex90 This problem started after an update for [`gulp-clone`](https://www.npmjs.com/package/gulp-clone) was released.

I made a pull request this morning with a fix so hopefully it gets merged. You can do a temporary fix by commenting out a line as @himanshu230 mentioned in his [comment](https://github.com/Semantic-Org/Semantic-UI/issues/6067#issuecomment-354995802).

fsworld009:
I manually changed gulp-clone to 1.0.0 and remove that line as @himanshu230 suggested but still cannot get any css file after `gulp build` command. Any idea?

Both Semantic UI (2.2.13) and gulp CLI (3.9.1) are the latest.

specimen151:
I have found that it might have something to do with the file size. At least these are my findings:

Editing semantic.json, the "components" array, and removing e.g. half of the components will build the css with  no problem. Test with `gulp build-css`. You can remove whichever components you want (removing one is not enough), so that's why I think it has something to do with the total size of if.

fsworld009:
@specimen151 I think it might also because you forgot to specify the theme for certain components (or you put an invalid theme name?) in `theme.config` but you included those components in "components" array, which causes the build script fail. At least for me I am able to build all components into one css file.

[...]

PR #6089, opened 2018-01-14
fix build/css.js, so gulp build can generate css files successfully

charlietag:
By default, "compressedStream" should be like a function. But it's not.
So I failed to build-css.
This way I can gulp build-css successfully.

Please take a glance.
Thanks a lot.
And appreciate for all your great work on Semantic-ui, I really like it !


jlukic:
See https://github.com/Semantic-Org/Semantic-UI/issues/6067 there was an issue upstream

PR #6110, opened 2018-01-20
Updated YouTube colour

hammy2899:
Updated the YouTube colour to the new red.

https://www.youtube.com/intl/en-GB/yt/about/brand-resources/


jlukic:
Very clearly wrong now, thanks!

Issue #5733, opened 2017-09-03
[Modal] .scrolling.content’s scrollbar can’t be dragged in Firefox (55.0.3)


ashkonBG:
In **Scrolling Content Modal**, the slider is not draggable to scroll but works fine with mouse wheel. If only you click and hold the slider then drag it to the middle of the screen, somehow you can scroll up and down. Still looking for the problem.

Banandrew:
Hi @ashkonBG, I couldn’t reproduce. What version of Chrome were you using when you noticed that?

ashkonBG:
Hey Andrew, there's no need to reproduce it just go to [here](https://semantic-ui.com/modules/modal.html), find the **Scrolling Content** press Run Code to launch the modal. Now if you drag the slider, it won't scroll unless dragged to the middle of screen.
Works fine with Chrome but not in Firefox (55.0.3).

Banandrew:
@ashkonBG Custom scrollbars apply to Webkit-based browsers only, so they don’t affect Firefox. I’ve just tested `55.0.3`, and it works as expected. What’s your OS?

ashkonBG:
Windows 10

ashkonBG:
@Banandrew i don't think this is a custom scrollbar, it's just a nested scroll. Maybe I am wrong, don't know.

Banandrew:
@ashkonBG Could you please post the modal’s markup that you’re using, the one with scrolling content?

ashkonBG:
Yes, the one says **Scrolling Content** with "New in 2.2.11" teal label in front of it.
Maybe I should make a video out of it!

Banandrew:
@ashkonBG When I asked for the modal’s markup, I meant the HTML that you’re using.

Banandrew:
@ashkonBG Nevermind, I understood what you said incorrectly.

ashkonBG:
@Banandrew like I said there's no need to see what I am using or how to reproduce the scenario. Even the official example in [here](https://semantic-ui.com/modules/modal.html#/definition) doesn't work.

Banandrew:
@ashkonBG Yes, sorry, it’s been a busy day, and my mind seems to be shutting off. In any case, I can’t see it on a very similar setup (also Windows 10), and we can’t fix it without some way of debugging it. Please record a video of it, and I’ll leave the issue opened in case there will be people that can confirm it.

[...]

PR #6132, opened 2018-02-02
Fix for #5733

ashucg:
There is an issue with scrolling content modal's scrollbar on some versions of Firefox and particular OS. To fix this issue, I have removed the perspective property for appropriate selector.


jlukic:
Perspective is necessary for 3d animations like `vertical flip` to render properly.

ashucg:
I think usability should be of higher priority than animations. Anyhow, thanks for your review

Issue #6135, opened 2018-02-05
Missing header size rule in `.tiny` modal size


icruces:
It seems a copy&paste or merging error. 

In `modal.less`, the rule:
```
/* tiny */
.ui.tiny.modal > .header:not(.ui) {
  font-size: @tinyHeaderSize;
}
```
is missing. And the small one is duplicated:
```
/* mini */
.ui.small.modal > .header:not(.ui) {
  font-size: @miniHeaderSize;
}
```
...
```
/* Small */
.ui.small.modal > .header:not(.ui) {
  font-size: @smallHeaderSize;
}
```

icruces:
I have created a pull request with a fix: #6137

lubber-de:
This is fixed in the community fork https://fomantic-ui.com

PR #6137, opened 2018-02-05
Fix for #6135 (missing modal `.tiny` header size)

icruces:


Bredo:
Why is the `.ui.small.modal > .header:not(.ui)` style removed?

icruces:
@Bredo The rule was duplicated. See line 500 of modal.less file.

lubber-de:
Fixed in [Fomantic-UI](https://fomantic-ui.com) 2.6.4

Issue #2846, opened 2015-08-14
Bold Font Weight Variable


aleesley:
Having a bold font weight site variable would be really useful for custom theme fonts with multiple weights. 

Example:
@boldFontWeight: 600;

vinayakkulkarni:
https://github.com/Semantic-Org/Semantic-UI/commit/faf0192241db8cb00874f5e2e13c458e987eb857#diff-09e5b8255f2710bb86860637e7e2c02aR17

Isn't it already implemented?

PR #6167, opened 2018-02-17
Add/utilize `fontWeight*` variables

jaridmargolin:
Right now, semantic-ui utilizes `normal` and `bold` values for setting font-weight's throughout the codebase. By default, browsers will map the `bold -> 700` and `normal -> 400` (this is somewhat of a simplification). For the most part, this strategy works.

I have, however, rubbed up against a requirement where I need to remap these values to different numerical weights (500 and 300). The `@font-face` declarations come down from a 3rd party service and I have no means of adjusting how they are defined. Right now, the only way I was able to accomplish this within semantic-ui, was to declare variables/overrides for all occasions where font-weight is set. It was a horribly tedious and the end solution feels very brittle as the code base changes (created nearly 30 files to accomplish this).

This PR adds support for dynamically setting the font weights used throughout semantic UI by defining three new variables, `fontWeightBold`, `fontWeightNormal`, and `fontWeightIcon`.

I believe this would properly address the concern brought up in #2846

---

As for testing these changes, I have built the dist files locally and manually previewed `examples/theming.html`. All UI appeared normal, however, I am uncertain if there is a more robust solution for verifying UI updates.


jaridmargolin:
@jlukic - I understand how difficult it can be to manage an open source project of this scale, so I hope nothing I say here comes off as offensive. I notice there is ~160 open PR's. I'm wondering what the likilhood is that this PR will get reviewed in a timely manner.

Is there anything I can do to help push it through? or at least get closure regarding wether or not there is interest in merging.

Thank you for you all of your hard work. Appreciate anything you can do here :)

jlukic:
This is pretty low friction update.

I think I'd prefer to do this as a single global variable `@bold` and `@normal` defined in `site.variables`, and not adding any additional variables.

To reconfigure on a component level you can just redefine `@bold: 600` inside the site or theme `button.variables` since all our variables are scoped **per** component.

jlukic:
Added in https://github.com/Semantic-Org/Semantic-UI/pull/6171

Issue #6145, opened 2018-02-07
[Dropdown] [v2.2.14] Multiple Dropdown pre-"selected" Options not visible


romanzipp:
When displaying a `<selection multiple>` with pre-selected Options, these options will not be displayed to the HTML.

**This bug is existing after upgrading to 2.2.14**

**Example - before JS initialized** (T1 will not be shown)
```html
<select multiple class="ui dropdown">
  <option selected>T1</option>
  <option>T2</option>
  <option>T3</option>
</select>
```

## 3/3 options selected
No Option gets displayed (selection active)

![2018-02-07 17_58_18-video 1 - pietsmiet admin opera](https://user-images.githubusercontent.com/11266773/35930251-b7659558-0c31-11e8-95d1-332fd69a39ce.png)

![2018-02-07 18_00_37-video 1 - pietsmiet admin opera](https://user-images.githubusercontent.com/11266773/35930254-b92ddf12-0c31-11e8-9b5d-4673b0e812da.png)

## 2/3 Options selected
Only the not-selected Option gets displayed and is selectable (selection active)

![2018-02-07 18_01_49-video 1 - pietsmiet admin opera](https://user-images.githubusercontent.com/11266773/35930258-bd39a172-0c31-11e8-9419-7bbe356752d2.png)

![2018-02-07 18_01_31-video 1 - pietsmiet admin opera](https://user-images.githubusercontent.com/11266773/35930261-bec8f2a4-0c31-11e8-979c-73ccf3ae77fb.png)

**When selecting T2, the expected behavior is shown until page refresh**

![2018-02-07 18_09_31-video 1 - pietsmiet admin opera](https://user-images.githubusercontent.com/11266773/35930424-256919ee-0c32-11e8-8229-0c14d75f20ff.png)

GammaGames:
#6123

camueller:
I noticed this problem as well. After downgrading to 2.2.12 pre-selected options are displayed properly!

qingquanwang:
:+1:

prudho:
Duplicate #6123

PR #6169, opened 2018-02-19
 Fix #6123, #6145

### LINKED ISSUE #6123: [Dropdown] Selected multiple options are hidden (2.2.14)
I've used to generate dropdowns from HTML in my forms, but Semantic 2.2.14 has a critical bug: selected options are hidden (but they are still considered as selected if you submit a form):

https://jsfiddle.net/3LaahL7e/1/

kjo6152:
add dropdown example.
check exists value using lable
- revert #3432
- It have a problem when initialize preselected options

### DIFF
--- a/examples/components/dropdown.html
+++ b/examples/components/dropdown.html
@@ -1,0 +1,123 @@
+<!DOCTYPE html>
+<html>
+<head>
+
+  <meta charset="utf-8" />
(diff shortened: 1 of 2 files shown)


kjo6152:
It has a problem when ignoreCase option is true, but preselected options have duplicate value.
I think It is not bug. It just user's mistake.

Issue #6109, opened 2018-01-20
Is Semantic UI development dead ?


GuacheSuede:
@jlukic  's last commit was 5/6  months ago and even then did not have many improvements nor much changes.

What would be the status of Semantic UI especially with Bootstrap having much more development than this ? Is it in active development or are any features planed for the near future ?

Note: This open source project goes dead without @jlukic or @quirkyjack , zero developement, zero updates, and for nearly a year at that. Even pull requests are not moving. Could there perhaps be a solution to this ?

Please do not say plans for something, for that might be for another half a year of waiting with no promises or answers

There will be a meeting at 6PM EST tomorrow on Hangouts !

**I hope to volunteer to maintain/run this project if @jlukic does not have the time for it**

arjarn:
Many many people are same interrogation : Semantic UI is great framework (with default but lot of good points). I'm really sad to see development stopped. For development (newer), using a framework without news about futur is not attractive.

etshy:
Same here I'd love to see semantic continue its development.
Maybe the semantic team works on something huge, like a V3 ?

GuacheSuede:
@etshy not in the Semantic-Org Repo, and @jlukic activity has been low too.

jeveloper:
Nothing since august last year

I've decided to move to material ui, this isnt worth investing time

I would however be happy to pay an annual fee for a premium plan that will guarantee bug fixes and improvements.

GuacheSuede:
@jeveloper I agree. perhaps a fork could be created

arjarn:
Agree too. I'm developper and do little stuff for fun. This month, I  begin new project (not for work) and have to choose between SUI and Metro-UI CSS V4 (I'm already using V3). Finally I choose Metro-UI : lot of high level components and very active developper.

Too bad : I like SUI, it's very good framework but no dev and with time, chatroom become empty... so no support.

For fork, possible if people are ready to participate...

GuacheSuede:
@arjarn Do you happen to know what is needed for Semantic UI to be better ?

arjarn:
In my opinion, SUI is almost perfect ^^
But :

  - be more easily customisable (skin) : actually it's difficult to modify color or add background on component (or I'm bad... possible because newer with SUI  ;-) )
  - several components missing (need 3rd party now) : date / time picker or have image slideshow.

But it's my feeling after my poor SUI experience, why don't making "BOX IDEA" here or elsewhere ?

[...]

PR #6182, opened 2018-02-21
Setup stalebot for stale issue management

levithomason:
This PR adds the config for [stalebot](https://github.com/probot/stale).  This was agreed in the [recent conversation](https://github.com/Semantic-Org/Semantic-UI/issues/6109#issuecomment-366872453) about how to help move things forward.

I've set issues to be marked stale at 90 days of no activity and closed if no activity persists for 30 more days.

We've been using stalebot with this config over at SUIR for a while and it is going well.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,22 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 30
+
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+
+# Label to use when marking an issue as stale
+staleLabel: stale
+
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed in 30 days if no further activity occurs.
+  Thank you for your contributions.
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,22 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 90
+
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 30
+
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
khornberg: Neither of these labels are in current use.
Perhaps `Confirmed Bug` would be a good starting label.

levithomason: Resolved, thanks!

jlukic:
Nice @levithomason

levithomason:
Thanks, I've enabled the bot: https://github.com/organizations/Semantic-Org/settings/installations/85175

hammy2899:
We might want to consider adding the `Enhancement` label to the exempt labels, lots of the issues which are being marked as stale labelled with enhancement are just getting bumped/+1.

levithomason:
This would defeat the purpose of the bot.  The (unfortunate) reality is, without action most ideas will never be implemented.  The purpose of the bot is to clean up those things that just won't actually ever happen.

hammy2899:
@levithomason Ye when I was looking through I kinda came to the same opinion, didn't realise there were so many issues labelled with Enhancement.

Issue #6186, opened 2018-02-21
Card content overflows the container


haikyuu:
Here is [a fiddle](https://jsfiddle.net/en2ou2mr/1/) reproducing the issue.

![screen shot 2018-02-21 at 11 55 42 am](https://user-images.githubusercontent.com/8558836/36478818-2dcacbec-16fe-11e8-804b-350c51cdc07d.png)

Thanks

hammy2899:
This is a pretty rare case for someone to have content with no spaces so I don't really see this being a problem. Unless you can provide an example of where this makes it an issue.

w96k:
Actually its easy to fix adding word-wrap CSS rule.
`word-wrap: break-word`

See fork on js fiddle: https://jsfiddle.net/kxvdb7Lx/3/
Would be nice to see it as a confirmed bug.

haikyuu:
> This is a pretty rare case for someone to have content with no spaces so I don't really see this being a problem. Unless you can provide an example of where this makes it an issue.

I believe things should be working by default. But this is your call as maintainers

lubber-de:
Prepared a PR for Fomantic-UI adopting the fix by @w96k here https://github.com/fomantic/Fomantic-UI/pull/703

coder8keyboard:
> This is a pretty rare case for someone to have content with no spaces so I don't really see this being a problem. Unless you can provide an example of where this makes it an issue.

E.g. file names.

PR #6188, opened 2018-02-21
Fix for #6186. Card word wrap.

w96k:
Fixed word-wrapping for long words without spaces (example: rsa key).


w96k:
up, its not a issue tho

PR #6189, opened 2018-02-21
Fix #6185.

### LINKED ISSUE #6185: [Modal] [v2.3.0] Multiple modals incorrect positioning
Hi,
It seems that in v2.3.0 was removed the fixed positioning for the modals, which leads to incorrect positioning when multiple modals is active.
semantic-ui com_modules_modal html
https://semantic-ui.com/modules/modal.html#/examples

w96k:
If default behavior is just modal on top of another modal then
everything works fine with 'position:fixed'.

### DIFF
--- a/src/definitions/modules/modal.less
+++ b/src/definitions/modules/modal.less
@@ -33,6 +33,7 @@
   transform-origin: @transformOrigin;
 
   flex: 0 0 auto;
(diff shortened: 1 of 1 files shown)


brodybits:
Still looks relevant

ref: #6185

PR #6192, opened 2018-02-21
Fix #6184. Centering icons.

### LINKED ISSUE #6184: Circular icons don't display properly in v2.3.0
There is a problem with icons with circular styles. Some icons are not centered in the circle.
image
The same problem with .circular.icon.button
image

w96k:
Demo: https://semantic-ui-icon-centering-btfdqouqqk.now.sh/


jlukic:
I've fixed this with variables only.

w96k:
Nice

PR #6202, opened 2018-02-22
fix name of 500px icon class

championswimmer:


hammy2899:
@championswimmer The 500px icon works as intended and the name is set to `\35 00px` because of the following reasons:

> In CSS, identifiers (including element names, classes, and IDs in selectors) can contain only the characters [a-zA-Z0-9] and ISO 10646 characters U+00A0 and higher, plus the hyphen (-) and the underscore (_); **_they cannot start with a digit_**, two hyphens, or a hyphen followed by a digit.

championswimmer:
How is having a space in the name permitted ? 
Most linters I use (and IDEs) are showing it as an error

hammy2899:
@championswimmer I use PhpStorm & WebStorm and they both show errors too but it still works, i'm pretty sure its just the way CSS escapes the character.

PR #6204, opened 2018-02-23
A fluid dropdown should be fluid

shadoWalker89:
When we have a `.fluid.dropdown` inside of `.inline.fields` the dropdown is set to `auto` because of the `.inline.fields`
A fluid dropdown should always be 100%

This is an example
The dropdown should take all the red space but it does not
https://jsfiddle.net/dq9urpzn/8/


lubber-de:
Fixed since [Fomantic-UI 2.4](https://github.com/fomantic/Fomantic-UI/releases/tag/2.4.0)
See https://jsfiddle.net/g40rd7b3/

PR #6211, opened 2018-02-23
Fix #5328, but for buttons

### LINKED ISSUE #5328: [Label] Icons should have correct margin when placed after the text
This works:  

```
<a class="ui label">
  <i class="mail icon"></i>
  Mail
</a>
```

This, almost works (the spacing is off):

```
<a class="ui label">
  Mail
  <i class="mail icon"></i>
</a>
```

Here is a screen shot of the difference:

image

Only if I use the delete icon is the spacing as it should be.  Here is the code:

```
<a class="ui label">
  Tag
  <i class="delete icon"></i>
</a>
```

Here is a screen shot showing the difference in the spacing:

image

Is there any way to get the spacing to work for other icons?  If so, can you please let me know how?  

Thanks.

w96k:
Demo: https://semantic-ui-icon-button-zvfzhufxuq.now.sh/

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -693,7 +693,7 @@
 .ui.icon.buttons .button > .icon,
 .ui.icon.button > .icon {
   opacity: @iconButtonOpacity;
-  margin: 0 auto;
+  margin: 0em auto;
 }
 
 .ui.right.icon.button > .icon {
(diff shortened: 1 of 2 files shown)


w96k:
up

jlukic:
Icon buttons are not intended to have text content. `icon button` is an "icon only button"

Issue #6221, opened 2018-02-27
[Dropdown] Category Search yields duplicate results


neokio:
When searching is set to use multiple fields (such as "category" and "title"), and when the search matches on both category AND title, a category search shows the same line twice.

![screen shot 2018-02-26 at 8 20 13 pm](https://user-images.githubusercontent.com/146623/36708972-ddc0c276-1b32-11e8-9f63-de5355df245c.png)

See [jsfiddle here](https://jsfiddle.net/f3wur1h8/6/).

Any tips on how to omit redundant results?

prudho:
`fullTextSearch` parameter set to `exact` seem to broke something while searching on multiple fields... (summon @hammy2899 and @jlukic).

Just initialize your search component with `fullTextSearch` set to `true` or `false`, according to the desired behavior.

jlukic:
Will address in patch this week.

jlukic:
Fixed in `2.3.1`

PR #6223, opened 2018-02-27
Fix #3486: [Search] add exactResults on duplicate check

### LINKED ISSUE #3486: A pointed tip should reflect color of active menu item
#### Steps to reproduce
1. Just open https://jsfiddle.net/z0vzw26e/3/
#### Expected

The pointed tip of an active item in the inverted menu should reflect the background-color of the the active item.
#### Actual result

The pointed tip is colored with black, which is a background-color of normal tab in inverted menu.

prudho:
### Closed Issues
#6221 

### Description
Avoid possible duplicate results when `fullTextSearch` is set to `exact`.


jlukic:
Looks great, thanks!

Issue #6158, opened 2018-02-11
[Search] Support disabled class


flashspys:
At the moment `.ui.search` does not support `disabled`. Search should act like a regular `.ui.input` if the `disabled` class is present.

PR #6225, opened 2018-02-27
[Search] add disabled style

prudho:
### Closed Issues
#6158

### Description
Allow the disabled state for the search component.


jlukic:
Needs docs, can add

PR #6226, opened 2018-02-27
[Icons] FA5 Report

hammy2899:
Reported FA5 using a new method so that the icons can use the FA5 font files. It also allows the ability to enable/disable solid, outline and brand icons so you can choose which icons you want in your theme. This method also eliminates the risk of bugs.

The method is using a much more reliable script to do the port and it does checks with the FA5 icon metafiles to include all icons with the correct Unicode characters. It also does a check to include all the old semantic aliases with much higher accuracy. It also builds to docs categories to the exact match of the FA5 website since it scrapes the site for which icons belong in which category.

I also added the new icons from yesterdays patch to FA5 5.0.7 ([medical](https://fontawesome.com/icons?d=gallery&s=regular,solid&c=medical&m=free) and [logistics](https://fontawesome.com/icons?d=gallery&s=regular,solid&c=logistics&m=free))


hammy2899:
This is the PR for the docs page https://github.com/Semantic-Org/Semantic-UI-Docs/pull/383

jlukic:
There's a merge conflict in your pull request

jlukic:
We cant remove any existing variables like `@fontName` this will break any custom themes.

jlukic:
I'll get this in, this weekend.

jlukic:
I've reworked this so as to not break any backwards compatibility.

Overview of changes
* Kept existing src/icon definition
* Renamed 'regular' to 'outline'
* Moved brand/outline to overrides
* Made `solid-icons`, `icons`
* Moved the `content:` definitions for outline to `importOutlineFont` conditional 
* Moved brand definitions into `importBrandFont` conditional

My largest concern with this refactor is making sure that people using custom fonts arent affected, and that they arent loading additional font files by default. This is mostly accomplished by moving the custom code over to the `overrides` file. I've also made sure to preserve the current `@src`  rules as they may be using these to define font locations.

PR #6181, opened 2018-02-20
Fixed the icon bugs raised in the issues

hammy2899:
Fixed exchange.alternate icon Fixes #6175
Fixed arrows.alternate.horizontal
Fixed arrows.alternate.vertical
Fixed arrow.alternate.circle.*
Fixed user.circle.outline Fixes #6176
Removed link and left linkify icon Fixes #6180
Bumped envelope unicode to stop conflict with resize.vertical

PR #6232, opened 2018-03-01
[Icons] Fix #6231 Legacy FontAwesome classes point to the wrong icon

### LINKED ISSUE #6231: [Icons] Icons removed in FontAwesome 5 are wrong
Some icons that were removed in FontAwesome 5 but were kept for compatibility reasons point to the wrong icons. Example:

**i.icon.resize.horizontal** has `content: "\f337"` which correlates to _i.icon.envelope.open.outline_.
Instead it shoud point to the new **i.icon.arrows.alternate.horizontal** which has `content: "\f697"`.

I didn't check all the old icons, but I guess that some others might be wrong, too.

NoisyFlake:
Fixes #6231 

Includes the following icons:

cloud.download
cloud.upload
disk.outline
exchange
heart.empty
hourglass.zero
protect
resize.horizontal
resize.vertical
star.empty


jlukic:
I've checked each of these icons against the next build and I believe they should all be fixed now. We've separated out the outline, brand, and solid fonts in #6181

albb0920:
I got missing icon after I upgrade from 2.2 to 2.3.2.
It seems like disk.outline in still missing.

Demo
http://jsfiddle.net/a9zvdw6j/10/

PR #6242, opened 2018-03-05
[Button] Fixes #5643 Primary & Secondary Inverted Buttons

### LINKED ISSUE #5643: [Button] - Inverted primary & secondary don't display properly
**Description:**
When creating a inverted button the coloured buttons (green, blue, yellow etc.) all work correctly but the primary and secondary buttons do not.

**Steps to recreate:**

1. Create button with `ui inverted primary button` or `ui inverted secondary button`

**Expected result:**
The button should be outlined with the primary or secondary colour and when the user hovers over it should be filled with the primary or secondary colour.

**Actual result:**
The button is filled with the primary or secondary colour and has a little corner border (very hard to see)

**Image:**
image

**Test case:**
https://jsfiddle.net/hammy2899/pc13yp7f/1/

hammy2899:
### Description
Primary and secondary inverted buttons display incorrectly

**Expected result:**
image

**Actual result:**
image

### Fixes Issues
#5643

### Testcase
https://jsfiddle.net/hammy2899/pc13yp7f/1/

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -329,11 +329,21 @@
        Inverted
 --------------------*/
 
-.ui.inverted.button {
+.ui.inverted.button:not(.primary + .secondary) {
   box-shadow: 0px 0px 0px @invertedBorderSize @white inset !important;
   background: transparent none;
   color: @white;
   text-shadow: none !important;
+}
+.ui.inverted.primary.button {
+  box-shadow: 0px 0px 0px @invertedBorderSize @primaryColor inset !important;
+  background: transparent none;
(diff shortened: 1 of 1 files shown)


jlukic:
I've added this feature using different code in https://github.com/Semantic-Org/Semantic-UI/commit/16e49424180363bbec3c6aa2994848daf83d483b

This correlates to how other `inverted` and `basic inverted` styles. 

They also use `lightPrimaryColor` and `lightSecondaryColor` which are the variations of primary / secondary that are intended for dark bgs. 

I've added the necessary corresponding state colors for these inverted colors.

Issue #4187, opened 2016-06-28
tooltip not fully visible


Maxhodges:
not sure if this is a bug, but maybe not the best demo of tooltip since the text can't all be seen

![image](https://cloud.githubusercontent.com/assets/957271/16428850/c0e8643c-3dae-11e6-9ff4-0ee3a4c04ef2.png)

dorsha:
I also had this issue:
![image](https://cloud.githubusercontent.com/assets/5429033/16430768/fd8ca3cc-3d85-11e6-9143-1f3e9fb684d4.png)

The left side bar is: `ui left vertical sidebar`
It reproduced also for top fixed menu (`ui top fixed menu`).

It is important to fix, since we don't one to handle and hack the tooltip style.

jlukic:
I'm guessing the tooltip doc demo issue is on mobile where there is no margin?

@dorsha If you want popup to be clipped by something that is not `body` use `boundary` setting and JS popups. CSS only popups cannot do any special logic for positioning, just animate in and out.

dorsha:
OK, make sense.

Maxhodges:
@jlukic was not on mobile but the browser was not full-screen so sidebar was hidden. Here's the context

![image](https://cloud.githubusercontent.com/assets/957271/16458116/f18191d6-3e58-11e6-8c96-1ef5bfa5094b.png)

PR #6249, opened 2018-03-07
[Popup] Fix a white space property in tooltip

kikunantoka:
### Closed Issues
#4187 

### Description
I found tooltip is not fully visible when tooltip has many contents. This is caused by selecting "nowrap" property as white-space in tooltip after element. So, I changed white-space property into "normal" from "nowrap".

#### Before
before

https://jsfiddle.net/ca0rovs3/271/

#### After
after

https://jsfiddle.net/ca0rovs3/272/

### DIFF
--- a/src/definitions/modules/popup.less
+++ b/src/definitions/modules/popup.less
@@ -107,7 +107,7 @@
   position: absolute;
   text-transform: none;
   text-align: left;
-  white-space: nowrap;
+  white-space: normal;
 
   font-size: @tooltipFontSize;
 


hammy2899:
Some people might want tooltips to be on a single line. I think maybe there should be an option to keep this behaviour like `data-single-line`

kikunantoka:
I agree your opinion. I think your idea is really good. Should I include a `data-single-line` option in this PR?

hammy2899:
Yes, go for it!

kikunantoka:
Ok. I try to add a `data-single-line` option later.

I'll send another PR for documentation if I try to add it to the doc.

kikunantoka:
I added a `data-single-line` option.

jlukic:
This might be more complicated than it seems. `nowrap` solves inadvertent text wrapping caused by certain css conditions for the activating element.

kikunantoka:
Thanks for your comment. I understand that.
If my commits might cause inconvenience, please close this PR.

jlukic:
I think we can just add a wrapping variation, and leave default behavior

allenwu1973:
Not sure is really related to this, but I find setting `white-space: pre` and using line breaks is quite helpful for long tooltips.

lubber-de:
I prepared a solution using an optional `fixed` class for Fomantic-UI which enables wrapping and uses the popup width boundaries here https://github.com/fomantic/Fomantic-UI/pull/1216

Issue #6147, opened 2018-02-07
ui.menu.secondary .item.link.disabled still changing state when hovered


awei01:
In a `ui.menu.secondary`, `div.item.link.disabled` and `a.item.disabled` change state when hovered. I would expect that the `:hover` pseudo-class should not work when hovered. This is the behavior for `ui.menu`

Please see example here: https://jsfiddle.net/awei01/xho6Lxv9/

awei01:
I believe this compiled CSS is missing `!important`

```

.ui.menu .item.disabled,
.ui.menu .item.disabled:hover {
  cursor: default;
  background-color: transparent !important;
  color: rgba(40, 40, 40, 0.3) !important;
}

```

PR #6268, opened 2018-03-12
[Menu] Disabled menu items still changing state on hover

tcmal:
### Closed Issues

 #6147

### Description

Previously, a disabled item in a ui.menu.secondary would still change state when hovered over.

This PR makes the relevant tags important, preventing this from happening

### Testcase

[Before](https://jsfiddle.net/awei01/xho6Lxv9/)

[Fixed](https://jsfiddle.net/tcmal/j4vh2o48/4/)


jlukic:
Looks good to me

PR #6281, opened 2018-03-19
[Modal] Fix #6198 - added check whether focused element is in a viewport

### LINKED ISSUE #6198: [Modal] Window is automatically scrolling up on modal window close
Hello,

JSFiddle example: https://jsfiddle.net/4qaxx4fr/

I've got an issue with modal windows. Please open JSFiddle demo, in Chrome for example, scroll down the result window to the bottom, make sure it's height is small enough, so you **cannot** see the whole calendar.

Click buttons **month** -> **week** -> **month** -> **week** (several times). Now try to create a event today at 11pm (just click there) - a modal window will appear -> press **Cancel**.

What I get now and what is wrong - browser window is scrolled up, until the Title of the calendar!

As you see - there is no my own libraries included, only basic code from fullcalendar/semantic-ui documentation.

Moreover, just after a page refresh everything is Ok, but then pressing **month** -> **week** -> **month** -> **week** enables this scrolling up.

---

As I see now after some exploring - when modal window is closing this is what's happening in Semantic-ui code:
```javascript
...
onComplete : function() {
  settings.onHidden.call(element);
  module.restore.focus();
  callback();
}
...
```
... **onHidden()** is fired and then you return focus to the last focused object. That's an issue. If I have a focused element somewhere in the beginning of a webpage, then scroll down, show a modal window and close it - browser is automatically scrolling up to the beginning of the page to that focused element.

I understand why would you need to set focus, but how do I prevent such behavior if I really need?

imilo:
I believe there is no need to restore focus to the invisible element when modal window is closed. Then there won't be case like https://github.com/Semantic-Org/Semantic-UI/issues/6198

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -66,6 +66,7 @@ $.fn.modal = function(parameters) {
         $allModals,
         $otherModals,
         $focusedElement,
+        $focusedElementVisible,
         $dimmable,
         $dimmer,
 
@@ -497,13 +498,21 @@ $.fn.modal = function(parameters) {
             ;
             if(!inCurrentModal) {
               $focusedElement = $(document.activeElement).blur();
+              
+              var
+                elementTop = $focusedElement.offset().top,
+                elementBottom = elementTop + $focusedElement.outerHeight(),
+                viewportTop = $(window).scrollTop(),
+                viewportBottom = viewportTop + $(window).height();
+
+              $focusedElementVisible = elementBottom > viewportTop && elementTop < viewportBottom;
             }
           }
         },
 
         restore: {
           focus: function() {
-            if($focusedElement && $focusedElement.length > 0) {
+            if($focusedElement && $focusedElement.length > 0 && $focusedElementVisible) {
               $focusedElement.focus();
             }
           }


hammy2899:
I think this should be optional since some people might want to go back to the focused element.

imilo:
This PR doesn't disable going to the focused element if it is visible. But if you want add a general flag to return or not to return to the invisible element - still good for me.

hammy2899:
Ye, I think it should be an option so it doesn't change any existing behaviour.

imilo:
If I add new setting to the `$.fn.modal.settings = {` at the end of the file - will I be able to configure it both globally with:
```javascript
$.fn.modal.settings.focusInvisible = false;
```
and for the specific modal window with:
```javascript
$('.ui.modal').modal({
  focusInvisible: false,
});
```
?
I ask because I'm not sure whether you set/append such settings somewhere else in other files.

hammy2899:
From my knowledge the settings which are passed when calling the module override the global settings which are defined at the end of the file so yes if you do

```javascript
$.fn.modal.settings.focusInvisible = false;
```

```javascript
$('.ui.modal').modal({
  focusInvisible: true
});
```

The modal with `.ui.modal` will be true but all others will be false.

imilo:
Thanks for the idea, I've added needed changes. Now it is optional and enabled by default.

imilo:
Hello!

Can somebody merge this PR?

Issue #6187, opened 2018-02-21
Dropdown upward direction with not enough space


mariolamacchia:
Hi there!

It looks like the dropdown's direction can currently be set to `auto` and `upward` only. It would be nice to have a setting that uses upward direction only if there is enough space on top. 

I see two alternatives:
- making `upward` work like `auto` but favoring upward direction instead of downward, OR
- adding other two possible settings: `downward`, that goes down no matter what, and `auto-upward`, which goes up only if there is enough space

What are your thoughts? I'd be happy to submit a PR.

hammy2899:
I think some people will benefit from this and if your willing to make a PR with this in I will be happy to review it for you!

PR #6286, opened 2018-03-19
[Dropdown] Add `auto-upward` direction

mariolamacchia:
### Closed Issue
https://github.com/Semantic-Org/Semantic-UI/issues/6187

### Description
With this PR is now possible to open the dropdown menu using, without forcing, the upward direction.
The possible values of the `direction` setting are now:
- `auto`: downward, if there is enough space
- `auto-upward`: upward, if there is enough space
- `downward`: downward, forced
- `upward`: upward, forced

### Testcase

https://jsfiddle.net/kda6z4br/

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -2405,15 +2405,15 @@ $.fn.dropdown = function(parameters) {
             }
           },
           direction: function($menu) {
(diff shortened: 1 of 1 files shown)


mariolamacchia:
Well, after one year I guess we aren't merging this one 🙂Closing

PR #6305, opened 2018-03-26
[Message] Added example page

tcmal:
### Closed Issues

None

### Description

Adds an example page in `examples/components/` for the message component.

### Testcase

[Mirror](https://jsfiddle.net/tcmal/ggn349fL/)

### DIFF
--- a/examples/components/message.html
+++ b/examples/components/message.html
@@ -1,0 +1,156 @@
+<!DOCTYPE html>
+<html>
+<head>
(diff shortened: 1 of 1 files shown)


hammy2899:
I would suggest putting this into a grid layout

tcmal:
Other than the message size examples, it is in a grid, I thought it would interfere with the message sizes though.

PR #6345, opened 2018-04-21
[Form] Add * required marker from the label itself

shadoWalker89:
### Description and TestCase
Before this PR in order to have the red `*` the `required` class must be added to the parent `.field` of the label.

This works fine until i have this case https://jsfiddle.net/kcqg1tz5/4/

When dealing with radio i cannot add the * marker to the label with this text `Signup for newsletter` because it does not live within a `.field`.

Also putting the `.required` on the `.inline.fields` will add the marker to the radios label elements which is not what i want.

With this PR adding the `.required` to the `label` itself will add the marker to that label only


lucasdcrk:
Should be commited IMO

PR #6401, opened 2018-05-21
Removed all redundant stuff like px, em, rem on all 0 values in less files

ForzaSFerrari:
Just removed all redundant stuff like px, em, rem on all 0 values in less files, where I hope, not to added any new errors

### DIFF
--- a/src/definitions/collections/form.less
+++ b/src/definitions/collections/form.less
@@ -49,12 +49,12 @@
 
 .ui.form .field:last-child,
 .ui.form .fields:last-child .field {
-  margin-bottom: 0em;
+  margin-bottom: 0;
 }
 
 .ui.form .fields .field {
   clear: both;
-  margin: 0em;
+  margin: 0;
 }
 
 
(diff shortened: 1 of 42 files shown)


src/definitions/collections/form.less
```diff
@@ -296,19 +296,19 @@
...
 .ui.form .field.field input:-webkit-autofill {
-  box-shadow: 0px 0px 0px 100px @inputAutoFillBackground inset !important;
+  box-shadow: 0 0 0 100 @inputAutoFillBackground inset !important;
   border-color: @inputAutoFillBorder !important;
 }
 
 /* Focus */
 .ui.form .field.field input:-webkit-autofill:focus {
-  box-shadow: 0px 0px 0px 100px @inputAutoFillFocusBackground inset !important;
+  box-shadow: 0 0 0 100 @inputAutoFillFocusBackground inset !important;
```
namnm: `100px` => `100`? Look like there will be a lot of error like this.

PR #6415, opened 2018-05-29
definitions: site: replace font-smoothing with -webkit-font-smoothing

karan9:
reference:
https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth


karan9:
but font-smoothing is not actually a property. then why would semantic want to keep it?

jlukic:
Prefixes are added by autoprefixer. If other vendors decide to adopt the proprietary syntax it could be accommodated by the prefixer.

PR #6426, opened 2018-06-10
[Search] Update for category search style and use FlexBox

hamid-ne:
## Update and optimize code in search.less
**Enhancements :**
- Use flexbox display on category search instead of table display
- Use the features of Less to better code management and better readability

### DIFF
--- a/src/definitions/modules/search.less
+++ b/src/definitions/modules/search.less
@@ -8,13 +8,12 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


HamidNE1:
Changes described.
please check it.

HamidNE1:
@hammy2899 has asked you to re-review.

Issue #6439, opened 2018-06-18
[Label] Basic label has same colored background since 2.3.2


j5k:
### Steps
https://semantic-ui.com/elements/label.html -> Basic

### Expected Result
up to version 2.3.1

### Actual Result
colored background, if color was defined

### Version
2.3.2

### Testcase
https://semantic-ui.com/elements/label.html -> Basic

ColinFrick:
Caused by https://github.com/Semantic-Org/Semantic-UI/pull/6220
`background-color: @basicBackground !important;` should most likely be `background: @basicBackground !important;`

miguelcaravantes:
same issue

adamkewley:
I'm also experiencing this issue on a few deployments. Seems `.ui.green.label`'s `background-color` is overriding `.ui.basic.label`'s background color. 

It's visible on the current SemanticUI documentation site (see `basic` section):

https://semantic-ui.com/elements/label.html

ppetrone86:
same issue. I installed semantic ui 2.3.2 via npm in my angular app

llturro:
@basicBackground equals to `none @white;` becoming a background-color **invalid property value**, caused by `none`.

Since specific attribute have precedence over combined, a possible solution would be adding a new variable for specific background attribute `@basicBackgroundColor: @white;` in **label.variables** and change

`.ui.basic.[color].label {
  background-color: @basicBackground !important;
  ...
}`

for

`.ui.basic.[color].label {
  background-color: @basicBackgroundColor !important;
  ...
}`

in **label.less**.

OwenPattison:
Same issue, tried to open a pr to fix this, access denied.

kaesar:
Same issue. thanks

jlukic:
I'll fix in patch

alvaro-canepa:
Same problem for v 2.3.3

fridzema:
Are there any updates on this? 
Really like to use the last release without manual hacks.

hammy2899:
@fridzema Fixed in FUI https://github.com/hammy2899/Fomantic-UI

hisuwh:
This is on your demo site guys.  This is embarassing get it fixed:
![image](https://user-images.githubusercontent.com/4470351/43969066-dd6a4152-9cc0-11e8-8077-0cab68688d0f.png)

james-ff:
This is still an issue in 2.3.3.
- Important text in basic labels with a color cannot be read!

[...]

PR #6440, opened 2018-06-18
[Label] Use basicBackground with background

ColinFrick:
Instead of background the property background-color was used.

Fixes Semantic-Org/Semantic-UI#6439


jlukic:
The variable is indeed `background` 

https://github.com/Semantic-Org/Semantic-UI/blob/master/src/themes/default/elements/label.variables#L106

kevinresol:
This fix is correct.

lubber-de:
Fixed in [Fomantic-UI](https://fomantic-ui.com) 2.6

PR #6447, opened 2018-06-21
[Dependency] Add compatibility with Less 3.x

sciyoshi:
### Closed Issues
https://github.com/Semantic-Org/Semantic-UI-LESS/issues/30

### Description
The two issues are related to commits in Less: less/less.js@9fd1ef5 and less/less.js@77fde7e. The first change causes variables to be undefined, the second change makes the `.loadFonts()` import the Google Fonts CSS incorrectly.

Any custom theme.config files will need to be updated to add the (multiple) option when including theme.less.

Re-opened from https://github.com/Semantic-Org/Semantic-UI-LESS/pull/44

CC @levithomason


levithomason:
@jlukic This fix is blocking some SUIR doc updates I'd like to make.  I've tested the fix here with Less 2.x and 3.x and all is well.

JonRowe:
Hi, is there a way to track when this support will be released?

levithomason:
@JonRowe @MaxMilton @stevewillard, I would suggest using a tool like http://greenkeeper.io to keep your dependencies up to date.  You'll simply get a PR every time one of your dependencies is updated.

JonRowe:
@levithomason Github dependencies also lets me do this, but I was hoping for a way to see announcement of releases so I can go unpin the relevant version of less etc

jlukic:
I'll be releasing it today in a patch.

JonRowe:
Still broken with less 3.0.4 and 3.5.2 for me

Issue #6454, opened 2018-06-26
[Search] - set value - can't call


MrMarci666:
Hey guys, I wanted to reset the Semantic UI search field, after the onSelect event called.

When I call the behavior, but it does nothing!

$('.ui.search').search('set value', '');
Other ways, it works, when I call this function from another source.
I don't really understand why, maybe there is a bug.

Thank you!

TEST IT HERE:
[https://jsfiddle.net/9jcj5va8/46/](https://jsfiddle.net/9jcj5va8/46/)

hammy2899:
This will be due to this line https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/search.js#L237

MrMarci666:
Yes! Is it legal to repair that myself? Or I should wait for the official solution?

hammy2899:
@MrMarci666 you are free to open a pull request to suggest a fix, please read the [contributing guide](https://github.com/Semantic-Org/Semantic-UI/blob/master/CONTRIBUTING.md) before submitting.

lubber-de:
Fixed in https://github.com/fomantic/Fomantic-UI 2.5.0

MrMarci666:
This seems to be fixed.

PR #6455, opened 2018-06-27
[SEARCH] - inputClear after selection

MrMarci666:
Hi All!

I made a solution for a bug.

You can find out the bug here:
[THE ISSUE HERE WITH EXAMPLE](https://github.com/Semantic-Org/Semantic-UI/issues/6454)

I added a new field for settings!
now it is: inputClear
It can be true or false.
if false: put the selected value to input.
if true: clear the input.
It doesn't give the focus back after click, that is annoying too, but easily you can do it with adding this to the initialization of the .ui.search element:

```				
onSelect: function(result, response) { 
     $(this).find('input').focus(); 
}
```

Now, the solution is here:
adding this to the settings:
[HERE](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/search.js#L1271)
``` 
  // after selection clearing the input field
  inputClear	 : false,
```

AND

adding this [HERE:](https://github.com/Semantic-Org/Semantic-UI/blob/master/src/definitions/modules/search.js#L237)
```               
 if (settings.inputClear == true) module.set.value('');
else  module.set.value(value);
```

It's quite simple. And now you can initialize your .ui.search element like this:
`$(.ui.search).search({ inputClear: true })`
And it will cleared always after a selection made.

or set that false, or just don't do anything, and the module will fill out the input field with the selected value! 👍 

Sorry, I am very new here, I don't know how to edit the modules correctly, if you help me out, that would be great! 💃

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "semantic-ui",
-  "version": "2.3.1",
+  "version": "2.3.2",
   "title": "Semantic UI",
   "description": "Semantic empowers designers and developers by creating a shared vocabulary for UI.",
   "homepage": "http://www.semantic-ui.com",


hammy2899:
The files changed aren't to do with the issue mentioned.

MrMarci666:
Yes, how to edit them, that is a mystery for me now. I edited the comment, see the solution is there.

PR #6505, opened 2018-07-25
Update nag.js to support Cookies instead of $.cookie

unostar:
jquery-cookie no longer maintained, superseded by JS Cookie: https://github.com/js-cookie/js-cookie


lubber-de:
We removed the general jquery plugin dependency in the community fork Fomantic-UI by
https://github.com/fomantic/Fomantic-UI/pull/1795

Issue #6470, opened 2018-07-07
[INSTALL] LESS errors on gulp build on fresh install


neokio:
### Steps

% cd semantic
% gulp build

### Result

Two errors:
1. `[Error: variable @googleProtocol is undefined in file /Users/n/webroot/semantic/src/definitions/globals/site.less line no. 25]`
2. `[Error: Operation on an invalid type in file /Users/n/webroot/semantic/src/themes/default/elements/label.variables line no. 133]`

### Complete Output

```
[16:47:23] Using gulpfile ~/webroot/semantic/gulpfile.js
[16:47:23] Starting 'build'...
Building Semantic
[16:47:23] Starting 'build-javascript'...
Building Javascript
[16:47:23] Starting 'build-css'...
Building CSS
[16:47:23] Starting 'build-assets'...
Building assets
[16:47:23] Created: dist/components/site.js
[16:47:23] Created: dist/components/site.min.js
[16:47:23] Created: dist/components/form.js
[16:47:24] Created: dist/components/form.min.js
[16:47:24] Created: dist/components/accordion.js
[16:47:24] Created: dist/components/accordion.min.js
[16:47:24] Created: dist/components/checkbox.js
[16:47:24] Created: dist/components/checkbox.min.js
[16:47:24] Created: dist/components/dimmer.js
{ [Error: variable @googleProtocol is undefined in file /Users/n/webroot/semantic/src/definitions/globals/site.less line no. 25]
  message:
   'variable @googleProtocol is undefined in file /Users/n/webroot/semantic/src/definitions/globals/site.less line no. 25',
  stack: undefined,
  type: 'Syntax',
  filename:
   '/Users/n/webroot/semantic/src/definitions/globals/site.less',
  index: 415,
  line: 25,
  column: 0,
  callLine: NaN,
  callExtract: undefined,
  extract: [ '', '.loadFonts();', '' ],
  lineNumber: 25,
  fileName:
   '/Users/n/webroot/semantic/src/definitions/globals/site.less',
  name: 'Error',
  plugin: 'gulp-less',
  showProperties: true,
  showStack: false,
  __safety: { toString: [Function: bound ] } }
[16:47:24] Created: dist/components/reset.css
[16:47:24] Starting 'package uncompressed css'...
[16:47:24] Created: dist/components/reset.min.css
[16:47:24] Starting 'package compressed css'...
[16:47:24] Created: dist/components/dimmer.min.js
[...]

PR #6512, opened 2018-07-31
[Dependency] Fix compatibility issue with less >= 3.5 due to math changes

sciyoshi:
A follow-up to https://github.com/Semantic-Org/Semantic-UI/pull/6447, there's a new change in less 3.5 for how calculations are handled. This PR avoids the current interpolation that's happening and simply uses `calc()`. The compiled CSS will be a bit different but should work the same way.

Closes https://github.com/Semantic-Org/Semantic-UI-LESS/issues/30, https://github.com/Semantic-Org/Semantic-UI-LESS/issues/46, https://github.com/Semantic-Org/Semantic-UI/issues/6470.

### DIFF
--- a/src/themes/default/collections/menu.variables
+++ b/src/themes/default/collections/menu.variables
@@ -321,7 +321,7 @@
 @tabularVerticalBackground: none @tabularBackgroundColor;
 
 @tabularFluidOffset: 1px;
-@tabularFluidWidth: ~"calc(100% + "(@tabularFluidOffset * 2)~")";
+@tabularFluidWidth: calc(100% + (@tabularFluidOffset * 2));
 
 @tabularActiveBackground: none @white;
 @tabularActiveColor: @selectedTextColor;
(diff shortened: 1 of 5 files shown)


sciyoshi:
cc @jlukic @levithomason

902Labs:
For whats it's worth, this fixes the bug - :LGTM:

daniel-cottone:
@jlukic @levithomason Can we get a review on this PR?

ColinFrick:
Applying this patch will break with the current build system (gulp-less@3.5.0):
`@tabularFluidWidth: calc(100% + (@tabularFluidOffset * 2));`
will output
`calc(102%)`

corydeppen:
Is there anything preventing the fix introduced in fomantic/Fomantic-UI#97 from being merged here?

jlukic:
I'm going to look into adding this as a patch. Got lost in the shuffle.

courdek:
Any idea on when this patch may find its way into the npm packages?

brendon:
Does this mean that building `Semantic-UI-Rails-LESS` against https://github.com/metaskills/less-rails which uses https://github.com/cowboyd/less.rb which seems stuck at `Less 1.7.0` will now fail?

Looks like a perfect storm of abandoned dependencies! :D

lubber-de:
Fixed in https://github.com/fomantic/Fomantic-UI 2.5.0

PR #6516, opened 2018-08-02
Fix for vertical alignment in tables

byte916:
In current version classes for vertical alignment are not working if they are used for a whole table

The example of the problem here https://jsfiddle.net/cws74Lpk/3/


lubber-de:
Fixed since 2.7.0 https://github.com/fomantic/Fomantic-UI/pull/320

Issue #222, opened 2013-10-09
[UI] Rich Text Editor


comtom:
It would be very nice & handy to have a full featured text editor as a component.

There are several examples, like CLEditor or TinyMCE. TinyMCE is one of the most popular, but there are some better like this one:

http://mindmup.github.io/bootstrap-wysiwyg/

it has some bugs but is great. I've encountred problems while pasting MS Word styled text, except that it works great.

salarmehr:
I would be nice to have a lightweight markdown editor.

rubydog:
@oknoorap Have you been able to get pen editor working? In my case the javascript seems to be working fine but it I can't see the tooltip when I select text in text area.

hubertguillemain:
Coming back on this topic: there are many Rich text editors out there, absolutely
I think the requirement is to maintain look and feel with the rest of Semantic UI (buttons, font, focus, ....)

Issue #333, opened 2013-11-07
[Comment] Two person conversation


deneuxa:
What do you think of a special rendering for a conversation between 2 people ?
It could be nice to alternate the placement of the avatars (left & right) in order to make a clear distinction between "my messages" and "my interlocutor's messages"

Just for illustration (not coherent with the SUI style) : 

![conversation](https://f.cloud.github.com/assets/359847/1491573/795f1f70-47b1-11e3-931a-c5c79ace10ae.png)

psmokotnin:
@jlukic is it any news?

psmokotnin:
we're still wait for this)

PR #6523, opened 2018-08-06
[Dropdown] Selection should be announced to screenreader

limograf:
made div holding current selection a live region so the selection is announced by screenreader. Would also live to label the created input that precedes it on search/select component (wherever that may be), which is currently silent and misleads the user.

✖ Multiple features in one PR

### Issue Titles

Use the format: [Component] Adds Support for Thing

For example: [Build Tools] Adds Source Map Support
Or: [Button] Fixes Inheritance for Red Basic Active State

### Closed Issues
#222 #333 #444

### Testcase

https://jsfiddle.net/ca0rovs3/

[Consider showing "fixed" case with your fiddle]()

You can link to your JS using https://rawgit.com/

### DIFF
--- a/src/definitions/modules/dropdown.js
+++ b/src/definitions/modules/dropdown.js
@@ -3873,16 +3873,16 @@ $.fn.dropdown.settings.templates = {
     ;
     html +=  '<i class="dropdown icon"></i>';
     if(select.placeholder) {
-      html += '<div class="default text">' + placeholder + '</div>';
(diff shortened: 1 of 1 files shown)


Atulin:
I think you should consider creating a PR on [fomantic/Fomantic-UI](https://github.com/fomantic/Fomantic-UI) if you have any enhancements in mind. SUI is a little bit on the dead side nowadays.

PR #6531, opened 2018-08-09
[Flag] add uk as a country code for united kingdom flag

PhilipGarnero:
It is pretty annoying having to maintain a map for gb -> uk conversion
I simply added the corresponding class.


jlukic:
Definitely is missing, added in `next`.

PR #6560, opened 2018-09-03
A segment in loading state should have a minimum height

brendon:
### Description

The minimum height could be a factor of the @loaderSize in order for the spinner to appear within the segment rather than overflowing it when it's empty or too small.

### Testcase

https://jsfiddle.net/ymn8w4uv/

[fixed](https://jsfiddle.net/sw69ouac/1/)


brendon:
Fixed in Fomantic.

PR #6567, opened 2018-09-06
[Site] fix font-smoothing

tobiaslins:
### Description
`font-smoothing` is not valid. There are only webkit and mozilla font smoothing.


lubber-de:
Fixed in Fomantic-UI by https://github.com/fomantic/Fomantic-UI/pull/563

Issue #4779, opened 2016-11-22
Add default variables for alternate grey spelling


iansoper:
I would be nice if there were aliases for the less variables for the alternate spelling of grey. Depending on where you live (like the US) you might spell it "gray". If the LESS themes had: `@gray: @grey`, `@lightGray: @lightGrey`, etc... that would be helpful.

levithomason:
Indeed, this correct.

>As the above chart indicates, _grey_ is much more preferred in British English.
>As the above chart shows, _gray_ is favored in American English.

http://writingexplained.org/grey-or-gray-difference

This gets me every time I add `gray` to a component and nothing happens :)

brodybits:
Fix is wanted ref: <https://github.com/Semantic-Org/Semantic-UI-React/issues/2574>. What is the status?

brodybits:
status please?

mcclowes:
Have made a PR: https://github.com/Semantic-Org/Semantic-UI/pull/6579

#6579

PR #6579, opened 2018-09-13
fix issue #4779 - grey vs gray variables and classes

mcclowes:
### Issue Titles
[Variables] Added @gray and associated variants
[Various] class `.gray` works just like `grey`

### Closed Issues
#4779 

### Description
Grey is the English spelling, but is not consistent with CSS. Therefore, to avoid confusion, gray should be default, whilst still supporting grey for existing users.


hammy2899:
This would create a breaking change, the best option here is to have `grey` and `gray` as options.

mcclowes:
I made them both options

mcclowes:
I did actually do that, it appears I just haven't built the css. Will do now.

mcclowes:
@hammy2899 Built. This should not be a breaking change now

hammy2899:
You have a merge conflict.

mcclowes:
Yes I see that. I didn't touch `RELEASE-NOTES.md`, nor do I know how it's supposed to be structured so perhaps you could advise on that.

Are you satisfied that that the change is no longer a breaking change though?

hammy2899:
Yes, the change is fine now.

Replace the release notes file with the one from the `next` branch that should solve the conflicts also try and remove the dist files from the commit.

mcclowes:
Thanks, got rid of a load of unintended diffs.

src/themes/default/elements/icon.overrides
```diff
@@ -36,11 +36,9 @@ i.icon.linkedin.in:before { content: "\f0e1"; }
 i.icon.zoom.in:before { content: "\f00e"; }
```
hammy2899: Why was this file edited?

src/definitions/modules/search.js
```diff
@@ -337,7 +337,7 @@ $.fn.search = function(parameters) {
               apiSettings = {
                 debug             : settings.debug,
                 on                : false,
-                cache             : true,
```
hammy2899: Why was this changed?

package-lock.json
```diff
@@ -1,102 +1,104 @@
 {
```
hammy2899: This file shouldn't be added

src/definitions/modules/search.js
mcclowes: I think the issue is I branched off master rather than next

package-lock.json
mcclowes: sure

src/themes/default/elements/icon.overrides
mcclowes: I think the issue is I branched off master rather than next

PR #6582, opened 2018-09-14
[Label] Basic background label definitions

lasley:
### Description

* The `@basicBackground` variable is set to `none @white`, which is meant to be used in the `background` CSS property of elements. Labels use the variable in `background-color` though, which creates an invalid rule. Switch to use `background` instead.

### Testcase

https://jsfiddle.net/5xyzedp0/1/

### DIFF
--- a/src/definitions/elements/label.less
+++ b/src/definitions/elements/label.less
@@ -566,7 +566,7 @@ a.ui.red.label:hover{
 }
 /* Basic */
 .ui.basic.red.label {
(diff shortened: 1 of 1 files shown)


ColinFrick:
Duplikate of https://github.com/Semantic-Org/Semantic-UI/pull/6440

lubber-de:
Fixed in [Fomantic-UI](https://fomantic-ui.com) 2.6

Issue #6587, opened 2018-09-18
[Modal] Fullscreen modal stuck to left edge of page


GammaGames:
### Steps
1. Create a fullscreen modal in html
2. Call the modal function on it and show

### Expected Result
Modal should be centered

### Actual Result
Modal is stuck to left edge of page

### Version
2.4.0

### Testcase
https://jsfiddle.net/Ljcyd5ek/

It can be fixed by adding the following css into the page to override it:

```css
.ui.fullscreen.modal {
  left: 2.5% !important;
}
```

phlfvry:
+1

phlfvry:
@GammaGames The CSS to override isn't working for me. I'm using the SUI CSS only version and even tried editing the source stylesheet file to no avail. Any other quick dirty fix?

GammaGames:
Yeah, the `.ui.fullscreen.scrolling.modal` was more specific, so it was overriding the base `.ui.fullscreen.modal` rules. If you want to read more, [here's](https://stackoverflow.com/a/25105841/3903479) a short explanation.

Also, in the future just add a thumb to the thread instead of a +1 comment. Everyone subscribed (with this thread it's not many) gets a notification whenever someone comments, and GitHub has tools to sort by the most reactions to see which issues have the most support.

jlukic:
I'll fix this.

SaphuA:
@jlukic 

This does not appear to be fixed in v2.4.2??

See -> https://codepen.io/anon/pen/aPJLya

lubber-de:
Fixed in https://github.com/fomantic/Fomantic-UI 2.6.1

PR #6602, opened 2018-09-26
Fullscreen modal aligned to left side of the screen without centering

beastie101:


lubber-de:
Does not work. See #6587 for correct fix

PR #6629, opened 2018-10-16
[Dropdown] Fixes unwanted text selection

chaaya:
### Closed Issues
None reported.

### Description
Text selection occurs when clicking the dropdown multiple times without leaving focus.
dropdown-selection

### Testcase
[See dropdown example](https://semantic-ui.com/modules/dropdown.html)
Click the dropdown down caret icon multiple times. It is also easily reproduced with the 
inline dropdown.


lubber-de:
Implemented since 2.6.4 https://github.com/fomantic/Fomantic-UI/pull/247
thanks @chaaya

Issue #6622, opened 2018-10-14
[Build] Failing silently in 2.4.1


eoliphan:
### Steps
run `gulp build`

### Expected Result
Output CSS bundle

### Actual Result
No CSS, the build process stops after creation of checkbox.min.css.  Reverting to 2.4.0 works with no issues
```
[09:35:55] Created: ../resources/public/css/components/accordion.min.css
[09:35:55] Created: ../resources/public/css/components/checkbox.css
[09:35:55] Created: ../resources/public/css/components/checkbox.min.css
Erichs-MBP:semantic ErichOliphant$ 
```

### Version
2.4.1

ColinFrick:
I can confirm this problem

himanshu230:
Hi All,
I have made a PR which solves the issue #6631

jlukic:
I've been on vacation in Ireland from about an hour after releasing `2.4.1` and now. 

Will take a look at issue and related PR

jlukic:
I'm unable to reproduce the issue. I'm currently able to build all css with `node -v` `8.9.0`

The steps I'm using to reproduce
1) Create new directory `fake/`
2) Run `npm install --save semantic ui`
3) `cd semantic` then `gulp build`
4) Confirm all files are in `dist/` and confirm version is `2.4.1`

jlukic:
Comments left for fix on #6631

jlukic:
`2.4.2` released with fix

PR #6631, opened 2018-10-17
Fixes the issue where gulp build was not building all css files #6622…

himanshu230:
### Issue Titles
[Build Tools] Fixes gulp build not building all css files

### Closed Issues
#6622 #6067 

### Description
Inside css.js gulp build file, `compressedStream` was set by piping `stream` to `clone()`. But later on, that variable was again set. It means that the initial` compressedStream` was left open, which caused backpressure and paused the original stream. As a result, all css files were not built. Solved the issue by removing the first open stream.

### DIFF
--- a/tasks/build/css.js
+++ b/tasks/build/css.js
@@ -79,12 +79,9 @@ module.exports = function(callback) {
     .pipe(flatten())
   ;
 
-  // two concurrent streams from same source to concat release
-  uncompressedStream = stream.pipe(clone());
-  compressedStream   = stream.pipe(clone());
-
   // uncompressed component css
-  uncompressedStream
+  uncompressedStream = stream
+    .pipe(clone())
     .pipe(plumber())
     .pipe(replace(assets.source, assets.uncompressed))
     .pipe(gulpif(config.hasPermission, chmod(config.permission)))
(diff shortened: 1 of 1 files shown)


charlycou:
I had to modify `print        = require('gulp-print'),`  to `  print        = require('gulp-print').default,` to make it work from my setup.

ccit-spence:
Tested and this fixed the issue I had with no css file being generated - V2.4.1

jlukic:
The compressed cloned stream was still referencing the original stream. With the gulp dependency updates this is causing issues.

```javascript
 compressedStream = stream
    .pipe(plumber())
   /* etc
```

it should just be 

```javascript
 compressedStream
    .pipe(plumber())
   /* etc
```

jlukic:
I'm cutting a new release now.

jtl385:
[Fixes the issue where gulp build was not building all css files #6622…](http://loopback.planatee.com:3000/c/xSvUwsT3/33-fixes-the-issue-where-gulp-build-was-not-building-all-css-files-6622)

jtl385:
[Fixes the issue where gulp build was not building all css files #6622…](http://loopback.planatee.com:3000/c/xSvUwsT3/33-fixes-the-issue-where-gulp-build-was-not-building-all-css-files-6622)

PR #6653, opened 2018-11-03
[Horizontal List] - icon not inline

zyzniewski:
An icon is not aligned properly within a horizontal list: https://jsfiddle.net/306w5p7g/

Regression introduced with: https://github.com/Semantic-Org/Semantic-UI/commit/6f2e1a85a07c198d13866dd1135677aa64583d07 - sets `.content` width to 100%

Solution: overwrite `.content` width for `.horizontal.list` https://jsfiddle.net/Lk1n7gwh/


lubber-de:
Fixed. See https://jsfiddle.net/Lk1n7gwh/1/

PR #6691, opened 2018-11-28
[Progress] Add properties for infinite loop as loader style

joonas-yoon:
Add settings for animation with infinite loop like loader.

I think it can be use instead of loader by who want progress-style-loader.

here is demo on JSFiddle:
https://jsfiddle.net/J00nas/xpvt214o/970756/


lubber-de:
We prepared a CSS only solution for Fomantic-UI here https://github.com/fomantic/Fomantic-UI/pull/735

Issue #6706, opened 2018-12-12
install fails with pnpm


cdfa:
✖ USAGE QUESTIONS → Use these dedicated resources:
      Docs - http://www.semantic-ui.com
      Chat - https://gitter.im/Semantic-Org/Semantic-UI
      SO - https://stackoverflow.com/questions/tagged/semantic-ui?sort=votes

✔ Enhancements → Be specific. Assume backwards compatibility is a necessity. Suggest implementation when possible.
✔ BUGS → ❤❤❤. Keep in mind some bugs may not be immediately fixable due to backwards compatibility or CSS limitations.

### Title (Put in field above)
Use the format: [Component] Component Should Do X
i.e. [Checkbox] onChange Should Fire When Update Triggered via DOM

### Description
I'm unable to install via [pnpm](https://pnpm.js.org/). It is 100% compliant with the npm module loading standard, but sometimes it breaks because packages do something outside of the standard. If you've looked into it and are confident this is a problem with pnpm, I'll report the issue there.

### Steps
- run `pnpm i semantic-ui`

### Expected Result
Command finishes without errors.

### Actual Result
```
Packages: +189
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Resolving: total 658, reused 658, downloaded 0, done
registry.npmjs.org/semantic-ui/2.4.2     |     install$ gulp install
registry.npmjs.org/semantic-ui/2.4.2     |     install:     at tryModuleLoad (module.js:506:12)
registry.npmjs.org/semantic-ui/2.4.2     |     install:     at Function.Module._load (module.js:
registry.npmjs.org/semantic-ui/2.4.2     |     install: Exited with 1
ERROR  semantic-ui@2.4.2 install: `gulp install`
Exit status 1
at Error: semantic-ui                                                                             at <anonymous>         /home/cdfa/.nvm/versions/node/v8.12.0/pnpm-global/1/node_modules/.registry.npmjs.org/pnpm/2.21.1/node_modules/pnpm/lib/node_modules/@zkochan/npm-lifecycle/index.js:301     var er = new Error('Exit…
[...]

PR #6719, opened 2019-01-03
fix: add less dependency for pnpm installs

DanielRuf:
✖ Multiple features in one PR

### Issue Titles

[dependencies] add less dependency for pnpm

### Closed Issues
#6706 

### Description
pnpm does and can not resolve dynamic require's

### Testcase

`pnpm i semantic-ui`

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "gulp-uglify": "^3.0.1",
     "gulp-watch": "^5.0.1",
     "jquery": "^3.3.1",
+    "less": "*",
     "map-stream": "^0.1.0",
     "merge-stream": "^1.0.0",
     "mkdirp": "^0.5.1",
--- a/pnpmfile.js
+++ b/pnpmfile.js
@@ -1,0 +1,14 @@
+module.exports = {
+    hooks: {
+        readPackage
+    }
+}
+
+function readPackage(pkg) {
+    if (pkg.name === 'accord') {
+        pkg.dependencies = {
+            less: '*'
+        }
+    }
+    return pkg
+}


DanielRuf:
You have to change the package.json of the `semantic-ui` package.
Also make sure to run `pnpm install` in `/home/cdfa/Projects/static-sticky/node_modules/.registry.npmjs.org/semantic-ui` or your folder.

DanielRuf:
I've done

```
pnpm i semantic-ui
cd /Users/druf/projects/pnpm-semantic-ui-test/node_modules/
cd semantic-ui
apply the changes
pnpm i
```

Or we could add it as peerdependency.

DanielRuf:
I've done

```
pnpm i semantic-ui
cd /Users/druf/projects/pnpm-semantic-ui-test/node_modules/
cd semantic-ui
apply the changes
pnpm i
```

cdfa:
Well, I installed your branch which includes the commit that adds less as a dependency and that did actually change something, because pnpm downloaded a new package. When this branch gets merged, people are just going to install it like i did, so not expecting any special setup. We should try to make it work that way too.

AFAIK, package's dependencies don't get installed in their respective 'node_modules' directories, but in the directory above (the user's application's `node_modules` directory), which is different from what you would get when running `pnpm i` in `node_modules/semantic-ui`.

DanielRuf:
Well, `gulp-less` already has `less` as dependency. `semantic-ui` too. So the issue is still `accord` I think as the dynamic require is still an issue.

I'll test further and maybe we have to fork `accord` or add some script or setting for pnpm.

DanielRuf:
https://github.com/pnpm/pnpm/issues/949

DanielRuf:
I've updated the PR.

cdfa:
I still get the same error...
Remember to use `pnpm i --force` to recreate the `node_modules` directory to clean any less package you might have installed otherwise.

DanielRuf:
Ah sorry, it is probably just used by the install command but not when it is part of a package.

cc @zkochan

DanielRuf:
> I still get the same error...
> Remember to use `pnpm i --force` to recreate the `node_modules` directory to clean any less package you might have installed otherwise.

I always create a new testfolder for local testing ;-)

DanielRuf:
> Ah sorry, it is probably just used by the install command but not when it is part of a package.

Ah yes it is.

DanielRuf:
Please try ```
module.exports = {
    hooks: {
        readPackage
    }
}

 function readPackage(pkg) {
    if (pkg.name === 'accord') {
        pkg.dependencies = Object.assign(
            {},
            pkg.dependencies,
            {less: '*'}
        )
    }
    return pkg
}
```

as pnpmfile.js in your project folder before installing / saving semantic-ui.

**Comment by DanielRuf** [comment]
```
pnpm install git://github.com/DanielRuf/Semantic-UI.git#fix/add-less-for-pnpm --save
Using hooks from: /Users/druf/projects/pnpm-test-4/pnpmfile.js
readPackage hook is declared. Manifests of dependencies might get overridden
Packages: +657
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Resolving: total 657, reused 657, downloaded 0, done
node_modules/.registry.npmjs.org/fsevents/1.2.4/node_modules/fsevents: Running install script, done in 357ms
node_modules/.github.com/DanielRuf/Semantic-UI/42dabde159bcac8c81d9dfee54d92baa0e32760f/node_modules/semantic-ui: Running install script, done in 2s
```

DanielRuf:
So my change is not needed. It is a change which is required at the level of `accord`.

I'll create a repo with the config.

DanielRuf:
https://github.com/DanielRuf/pnpmfiles/blob/master/semantic-ui/pnpmfile.js

cdfa:
It seems adding `less` as dev dependency as suggested [here](https://github.com/pnpm/pnpm/issues/762) also works, which seems like a better solution to me. Thanks for helping anyway!

ndc411:
first, i used gulp-less 4.x, but error was showm. then downgrade to 3.x,it works normally!

npm install --save gulp-less@3.5.0

PR #6731, opened 2019-01-14
[Button] style should not depend on order of class

### LINKED ISSUE #6730: styles should not depend on order of class
### Steps
https://jsfiddle.net/y8L6h5c9/

### Expected Result
Both button should float right.

### Actual Result
If `right floated` is not appeared in class exactly same order,
float is not working.

Toggling class in devtool breaks button's float state

### Version
2.4.2

### Testcase
Please see Steps.

yskkin:
### Closed Issues
closes #6730 

### Description
With current implemention, toggling class in devtool broke float state

### Testcase

[Show before with this fiddle](https://rawgit.com/)

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -607,14 +607,14 @@
        Floated
 --------------------*/
 
-.ui[class*="left floated"].buttons,
-.ui[class*="left floated"].button {
+.ui.left.floated.buttons,
+.ui.left.floated.button {
   float: left;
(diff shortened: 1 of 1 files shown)


hammy2899:
This is a design choice not a mistake. Also changing this will break things like the following
```html
<button class="ui left labeled right floated icon button">
  <i class="left arrow icon"></i>
  Next
</button>
```

https://jsfiddle.net/y1qp0vLk/

PR #6737, opened 2019-01-22
fixed submitting

0xflotus:


exoego:
Thanks your work!
Since Semantic-UI is inactive as of toda, I'd recommend adding a PR to [Fomantic-UI](https://github.com/fomantic/Fomantic-UI) as well.
Better chance of it being actually merged,

PR #6766, opened 2019-03-04
[Button] Use correct classnames in adding margin-left

dylmye:
### Closed Issues
n/a

### Description
Previously, the following line was copy-pasted in each button style section:
```css
.ui.buttons:not(.vertical) > .basic.primary.button:not(:first-child) {
  margin-left: -1px;
}
```
(copied from the static semantic-ui-css)

however, it appears that the button style (primary, secondary, positive, negative) was not updated in each usage. I've changed this so now this rule is applied to the style of the section the rule is contained in.

I hope this makes sense - I've also foregone the testcase section as this is such a minor fix.

### Testcase
n/a

### DIFF
--- a/src/definitions/elements/button.less
+++ b/src/definitions/elements/button.less
@@ -3263,7 +3263,7 @@
   box-shadow: 0px 0px 0px @basicColoredBorderSize @secondaryColorDown inset !important;
   color: @secondaryColorDown !important;
 }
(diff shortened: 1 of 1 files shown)


lubber-de:
Good catch!
I adopted your fix for upcoming Fomantic-UI 2.7.3 by https://github.com/fomantic/Fomantic-UI/pull/537

dylmye:
Ok to close this?

PR #6784, opened 2019-04-02
[Table] Cursor pointer for selectable rows or cells

zediogoviana:
### Description
Make cursor pointer visible when "selectable" prop is defined. It is visible in List but not on Table, allowing a more intuitive clickable component.

### DIFF
--- a/src/definitions/collections/table.less
+++ b/src/definitions/collections/table.less
@@ -500,21 +500,25 @@
 .ui.table tbody tr td.selectable:hover {
   background: @selectableBackground !important;
   color: @selectableTextColor !important;
+  cursor: pointer;
 }
 .ui.selectable.inverted.table tbody tr:hover,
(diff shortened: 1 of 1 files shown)


lubber-de:
@zediogoviana Thanks for the hint. I collected and shortened your changes for Fomantic-UI here
https://github.com/fomantic/Fomantic-UI/pull/611
See: https://jsfiddle.net/zgdjnvb5/1/

src/definitions/collections/table.less
```diff
@@ -500,21 +500,25 @@
...
+  cursor: pointer;
 }
 .ui.table tbody tr td.selectable > a:not(.ui) {
   display: block;
   color: inherit;
   padding: @cellVerticalPadding @cellHorizontalPadding;
+  cursor: pointer;
 }
 
 /* Other States */
```
MaherSaif:
```suggestion
/* Other States */
```

PR #6793, opened 2019-04-17
Fix search when item contains accents

zeden:


ColinFrick:
@zeden We fixed this in the community fork Fomantic-UI version 2.7.2:
https://github.com/fomantic/Fomantic-UI/pull/422

Additionally Internet Explorer does not support `normalize`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#Browser_compatibility

PR #6833, opened 2019-06-21
[Modal] Tears down mutation observer on modal destruction

### LINKED ISSUE #6832: [Modal] destroy should tear down mutation observer
### Steps

- Show a modal with `{observeChanges: true}`
- Hide the modal (either programmatically or by, e.g. clicking on the dimmer)
- Some DOM updates are performed during the hide/destruction phase of the modal

### Expected Result

Scroll lock is removed from the dimmer root

### Actual Result

Scroll lock is incorrectly reinstated after tear down of the modal

### Version
2.4.0

### Testcase

The test case is quite tricky to extract as the specific problem I'm facing is caused by the interaction of Angular and the teardown mechanism of the modal. I can make a PR that fixes the behaviour as I have patched my local version to achieve the correct result.

euangoddard:
Fixes #6832

### DIFF
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -138,7 +138,7 @@ $.fn.modal = function(parameters) {
         },
 
         destroy: function() {
-          if (module.observer) {
+          if (observer) {
             observer.disconnect();
           }
           module.verbose('Destroying previous modal');
@@ -163,7 +163,6 @@ $.fn.modal = function(parameters) {
               subtree   : true
             });
             module.debug('Setting up mutation observer', observer);
-            module.observer = observer;
           }
         },
 
--- a/src/definitions/modules/modal.js
+++ b/src/definitions/modules/modal.js
@@ -138,6 +138,9 @@ $.fn.modal = function(parameters) {
         },
 
         destroy: function() {
+          if (module.observer) {
+            observer.disconnect();
+          }
           module.verbose('Destroying previous modal');
           $module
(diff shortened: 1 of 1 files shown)


src/definitions/modules/modal.js
```diff
@@ -138,6 +138,9 @@ $.fn.modal = function(parameters) {
         },
 
         destroy: function() {
+          if (module.observer) {
```
lubber-de: `if (observer)` should do it alone, because this variable is already declared here https://github.com/Semantic-Org/Semantic-UI/blob/0a42dcdcb400d6fc275b04d6cb052a48efd68536/src/definitions/modules/modal.js#L79

src/definitions/modules/modal.js
```diff
@@ -160,6 +163,7 @@ $.fn.modal = function(parameters) {
               subtree   : true
             });
             module.debug('Setting up mutation observer', observer);
+            module.observer = observer;
```
lubber-de: As mentioned above, `observer` is already declared as private module variable and set by https://github.com/Semantic-Org/Semantic-UI/blob/0a42dcdcb400d6fc275b04d6cb052a48efd68536/src/definitions/modules/modal.js#L157, so this line should not be necessary

euangoddard:
@lubber-de thanks for the feedback - I'd missed that declaration and have changed in line with your suggestion

jlukic:
This looks good to merge.

Issue #4727, opened 2016-11-09
npm install -> gulp install error


abbiento:
I used the command npm install --save semantic-ui but after some download I receive this error in the file error.png. I've attached npm-debug.txt. How can I resolve the problem?

[npm-debug.txt](https://github.com/Semantic-Org/Semantic-UI/files/581307/npm-debug.txt)

![error](https://cloud.githubusercontent.com/assets/6938151/20145458/7def49ba-a6a0-11e6-853a-ab0b449ca16d.png)

webdesignberlin:
Hi.

I've tried to track down your problem. No problems with Windows 10, Windows 7 with node v4.6.1 and Gulp CLI version 1.2.2 and Gulp Local version 3.9.1.

Which Gulp and Node version do you use?

Best regards, Micha

wolfadex:
I'm on Ubuntu 14.04, node 6.6.0, Gulp CLI version 1.2.2 and Gulp Local version 3.9.1.

Could it be an issue with [accord](https://github.com/jescalan/accord)? I noticed that is the module that's throwing the error and on their github page their current build is failing.

wolfadex:
Also of note, this issue doesn't occur with bower.

abbiento:
Gulp CLI global, version 3.9.1
Npm global, version 3.10.9

I should ever install a local version of gulp? When I've tried to install gulp locally (same version 3.9.1) I've got the same error..

abbiento:
This are step for semantic-ui installation on my windows 10:
1) md sui
2) npm init (with all default)
3) npm install --save-dev gulp
4) npm install --save semantic-ui

and this is the error:

![error](https://cloud.githubusercontent.com/assets/6938151/20208466/ac9fd87a-a7ee-11e6-9c11-95fb1ac89384.png)

abbiento:
No resolution??

liam-cameron:
+1 exact same issue

liam-cameron:
Liams-MBP:project liamcameron$ npm install semantic-ui
npm WARN excluding symbolic link tests/withsymlinks/.hidden -> ../shown/.hidden
npm WARN excluding symbolic link tests/withsymlinks/bar.txt -> ../shown/bar.txt
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
[...]

PR #6853, opened 2019-08-02
:bug: Fixes docs build error from Node's fs

whoisryosuke:
`fs` from Node now requires a callback for the third parameter, which was missing from the docs Gulp build task file, causing the `gulp build-docs` process to break with the following error:

```
[15:59:18] Starting 'build-docs'...
Building Metadata
Copying examples
Copying LESS source
Building Semantic for docs
(node:72733) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
fs.js:113
      throw err;  // Forgot a callback but don't know where? Use NODE_DEBUG=fs
      ^

Error: ENOENT: no such file or directory, open '/Users/username/Development/HTML/docs/out/metadata.json'
ITs-MacBook-Pro-6:ksc-ui photoroom$ ./node_modules/gulp/bin/gulp.js build-docs
```

Adding a simple error callback (with optional console log for error reporting) **fixes this issue**. Running the docs build process now works.

Not sure if this has already been hashed out for the next release cycle.

✖ Multiple features in one PR

### Issue Titles

Use the format: [Component] Adds Support for Thing

For example: [Build Tools] Adds Source Map Support
Or: [Button] Fixes Inheritance for Red Basic Active State

### Closed Issues
#222 #333 #444

### Testcase

https://jsfiddle.net/ca0rovs3/

[Consider showing "fixed" case with your fiddle]()

You can link to your JS using https://rawgit.com/

### DIFF
--- a/tasks/docs/build.js
+++ b/tasks/docs/build.js
@@ -80,7 +80,10 @@ module.exports = function(callback) {
   gulp.src(config.paths.template.eco + globs.eco)
     .pipe(map(metadata.parser))
     .on('end', function() {
-      fs.writeFile(output.metadata + '/metadata.json', JSON.stringify(metadata.result, null, 2));
+      fs.writeFile(output.metadata + '/metadata.json', JSON.stringify(metadata.result, null, 2),
+        error => {
+          console.log("metadata failed", error);
+        });
     })
   ;
 


danShumway:
Are there any plans to merge this? We're trying to decide the best way to handle the current Semantic UI build issues (#4727). If the repository isn't being maintained any more at this point, it would be good to know so we can vendor the dependency and start applying our own patches.

PR #6876, opened 2019-09-23
[Theme]: allow null @site

cruzdanilo:
### [Theme]: allow null @site

allow "null" values for @site theme config variable, like all other variables already allow.
as the default theme is always imported, using "null" values for all variables decreases dramatically the bundle size by avoiding duplicates when using default theme. this improvement is noticeable even when using optimizers like cssnano.

ps: i am calling "null" values that cause the optional import not to find any file.


jlukic:
This looks good to me.

PR #6896, opened 2019-11-11
Remove pointer events for disabled toggle

tenKinetic:
Set pointer-events to none when checkbox is in toggle mode and is disabled

### Description

Came across this while developing a UI and didn't see any related issues.
A very simple update that just ensures when a checkbox is in toggle mode and is also disabled, the toggle cannot be manipulated.

### DIFF
--- a/src/definitions/modules/checkbox.less
+++ b/src/definitions/modules/checkbox.less
@@ -572,6 +572,11 @@
   background-color: @toggleOnFocusLaneColor !important;
 }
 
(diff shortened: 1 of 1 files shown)


lubber-de:
I adopted your PR for Fomantic UI by https://github.com/fomantic/Fomantic-UI/pull/1245
In fact all variants of checkbox, not only toggle, are affected

jaydeepkhatri (approved):

PR #6900, opened 2019-11-21
[Form] Resolved query 'set errors' leaving DOM element invisible.

akotulu:
Reported this issue with patch file, but it seems to be deleted by github or some other behind the scene force. Here is the fix. Currently adding errors to form via 'set errors' will leave the dom element invisible. Fix is to set the css class after error message.


hammy2899:
@akotulu This will most likely never be merged, its over 1 year now since last release of SUI (https://github.com/Semantic-Org/Semantic-UI/releases) take a look at [Fomantic-UI](https://github.com/fomantic/Fomantic-UI) the community fork.

lubber-de:
I cannot seem to create a usecase where your described scenario is happening.
See https://jsfiddle.net/uLrm0wcd/
`set errors` show the errors just fine... 🤔 
Could you please adjust the example, so we can see what exactly your code is going go fix?
In Fomantic-UI 2.7.3 we fixed a similar issue for safari not showing the message via CSS by https://github.com/fomantic/Fomantic-UI/pull/558

Issue #5745, opened 2017-09-05
[Form Validation] should not throw exception on `remove field( ... )` behaviour


cueedee:
Please see this [jsfiddle](http://jsfiddle.net/cueedee/wr2v8opt/).

Invoking `$( '.ui.form' ).form( 'remove field( field )' )` for an input field that has no validation rules set, throws an exception.

The expected behaviour is for it to be a no-op.

cueedee:
@stale-bot; Given the "Enhancement" status, I would presume that they still intend to address this issue at some point.

cueedee:
@stale-bot; this i getting annoying. If it's not `closed` it's not closed.

jameschenjav:
This is a bug, not an enhancement. Surprised no one fixed it in 3 years.

lubber-de:
As i [mentioned in your PR](https://github.com/Semantic-Org/Semantic-UI/pull/6940#issuecomment-588087079) this was already fixed in https://fomantic-ui.com last year. I wasnt aware this issue exists, otherwise i would have linked it before

PR #6940, opened 2020-02-19
[Form] Fixes "remove rule(field)" throw exception if field has no rules

jameschenjav:
### Closed Issues
#5745

### Description
`remove rule(field, rules)` should remove all rules when argument `rules` is omitted. But `field` may not have any rules, and it would raise exceptions because of `validation[field].rules = []`.

There is some guard code to check if `validation[field].rules` exists, but the order is wrong apparently. 

### Testcase
https://jsfiddle.net/whd3o60j/3/

Click "Remove All Validations", raised exception `Uncaught TypeError: Cannot set property 'rules' of undefined` (in Chrome)

### DIFF
--- a/src/definitions/behaviors/form.js
+++ b/src/definitions/behaviors/form.js
@@ -749,19 +749,19 @@ $.fn.form = function(parameters) {
 
         remove: {
           rule: function(field, rule) {
-            var
-              rules = $.isArray(rule)
-                ? rule
-                : [rule]
-            ;
+            if(validation[field] == undefined || !$.isArray(validation[field].rules)) {
+              return;
+            }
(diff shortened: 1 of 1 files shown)


lubber-de:
This is fixed in https://fomantic-ui,com
See your same fiddle using fomantic here
https://jsfiddle.net/x6h0bLrs/

jameschenjav:
hi @lubber-de 
i noticed semantic-ui repo stopped updating after i created this PR and mentioned it to our dev team. 
i suggested to our team after saw your comment, and today we've decided to adopt fomantic-ui for further upgrades. 
thanks for the great work.

